PR #2, opened 2022-07-15
docs: new readme

vladfrangu:


README.md
```diff
@@ -1,194 +1,64 @@
...
-    // Choose the first URL to open.
-    await requestQueue.addRequest({ url: 'https://www.iana.org/' });
+```typescript
+import { Actor } from 'apify';
+import { PlaywrightCrawler } from 'crawlee';
 
-    const crawler = new Apify.PlaywrightCrawler({
-        requestQueue,
-        handlePageFunction: async ({ request, page }) => {
+await Actor.main(async () => {
```
B4nan, line 44: We should probably prefer the `init/exit` way instead of `Actor.main`. I'd be in favour of having both examples, as people are familiar with the `main` method, but with `init/exit` you dont need to nest anything which is better for readability.

README.md
```diff
@@ -1,194 +1,64 @@
-# Apify SDK monorepo
+# Actor SDK
```
B4nan, line 1: this will be monorepo for actor sdk + more (apify-cli, apify-client, storage-local, ...)

vladfrangu: Oof, right... then we cannot share a readme for all of them (or we *can* but it'd be noisy for each module). Do you want me to instead make a readme in the apify module?

B4nan: We can deal with that when we add more packages, right now we can use the root readme.

vladfrangu: Well if I add monorepo to the readme that won't make much sense on npm's side... Your call, I can add it back

B4nan: Yeah I guess we should have separete readme per package here, still the root one can talk about actor sdk in more detail than just "this is a monorepo, actor sdk is one of the packages", especially now :]

vladfrangu: Then should I copy this current readme into the apify module and write up an actor sdk monorepo readme?

B4nan: Yop. Btw the root one is beging copied via `copy.ts` script.

PR #26, opened 2022-08-08
feat: add playwright scraper

B4nan:


B4nan:
cc @barjin, maybe you can reuse something from this

AndreyBykov:
@B4nan could not add you as a reviewer 🤷‍♂️

Updated the readme (as well as quite a few other things), tried to build on the platform - seems like all is good. There are some rough things to be polished in scraper tools, but seem to be working fine.

Using big playwright image and runs both chromium/firefox. Webkit has some issues with fingerprints for some reason. I guess if it's gonna be slow - in the future we could create Playwright scraper+chromium and separate with firefox one.

AndreyBykov:
btw - I put some links to Apify SDK with `sdk.apify.com` so that we would not need to return to it, hopefully, I used the correct links (but I'll check it of course).

packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
```diff
@@ -0,0 +1,277 @@
...
+            ],
+            "enumTitles": [
+                "Chromium",
+                "Firefox"
+            ],
+            "prefill": "chromium",
+            "default": "chromium"
+        },
+        "useChrome": {
+            "title": "Use Chrome (could only be used if Chromium Browser is selected)",
```
barjin, line 118: Maybe too long for a title? But the explanation in parentheses is absolutely crucial too.

packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
```diff
@@ -0,0 +1,277 @@
...
+            "type": "integer",
+            "description": "Maximum time the scraper will wait for the page function to execute in seconds.",
+            "minimum": 1,
+            "default": 60,
+            "unit": "secs"
+        },
+        "waitUntil": {
+            "title": "Navigation wait until",
+            "type": "array",
+            "description": "The scraper will wait until the selected events are triggered in the page before executing the page function. Available events are <code>domcontentloaded</code>, <code>load</code> and <code>networkidle</code> <a href=\"https://playwright.dev/docs/api/class-page#page-goto-option-wait-until\" target=\"_blank\">See Playwright docs</a>.",
```
barjin, line 211: `networkidle2` (default prefill) is not listed here. Btw could this be a dropdown? Idk the input schema that well though :)

barjin (commented):
thank you, great job :)

packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
AndreyBykov: It looks ok as it's just a boolean, and there's a whole screen empty) 👌

packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
AndreyBykov: Playwright does not support networkidle2, it has only networkodle, load, and domcontentloaded 🙂

B4nan: yes, its not supported, and you still have it as the default/prefill

image

AndreyBykov: Oh damn 🤦‍♂️

AndreyBykov: Fixed, as for the dropdown, actually yes, puppeteer could have an array of events there, while playwright support only 1 - will update.

packages/scraper-tools/src/browser_tools.ts
```diff
@@ -1,10 +1,18 @@
 import { Actor } from 'apify';
 import log from '@apify/log';
-import type { Page } from 'puppeteer';
+import type { CommonPage } from '@crawlee/browser-pool';
 import { inspect } from 'util';
 import { RESOURCE_LOAD_ERROR_MESSAGE, SNAPSHOT } from './consts';
 import { createRandomHash } from './tools';
 
+// TODO
```
B4nan, line 8: TODO what? lets comment whats missing

packages/scraper-tools/src/browser_tools.ts
```diff
@@ -100,6 +108,7 @@ export interface DumpConsoleOptions {
  * Chromium messages, usually internal errors, occuring in page.
  */
 export function dumpConsole(page: Page, options: DumpConsoleOptions = {}) {
+    // @ts-expect-error Parameter 'msg' implicitly has an 'any' type. // TODO
     page.on('console', async (msg) => {
```
B4nan, line 112: just add the type (ideally put there something more strict than `any`)
```suggestion
    page.on('console', async (msg: any) => {
```

packages/scraper-tools/src/browser_tools.ts
AndreyBykov: Pretty much. It's here because it was puppeteer only, now it has to support both. I just wasn't sure if that was exactly the right thing to do. Will remove todo

packages/scraper-tools/src/browser_tools.ts
AndreyBykov: it would work for example with JSHandle, but it would bring the puppeteer dependency, while we don't really want that here :/

B4nan: i dont care much about the type, this is more about "lets not overuse ts-ignore comments". i am fine with `msg: any`, more than with the ignore comment :]

packages/scraper-tools/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@apify/scraper-tools",
-    "version": "1.0.0",
+    "version": "1.0.1",
```
B4nan, line 3: why? this will release canary 1.0.2, are you aware of that?

AndreyBykov: But I just told you on Slack that once I bumped the version to 1.0.1 I was able to install the stuff locally - I thought you kinda confirmed that if it helps - to do so. Ok, reverting 👍

(although now events test does not pass for no reason).

AndreyBykov:
@B4nan so, apparently I fucked up package-lock on merging changes from master here: https://github.com/apify/apify-sdk-js/pull/26/commits/bff9f9e8e5ba164bb34de7134a87929dfa9479fa
Then I guess I re-generated it here https://github.com/apify/apify-sdk-js/pull/26/commits/3716dc3de2d469a542c42b3867f85a52d1d3f7f4 (not sure really, as tests were failing because of the version first, then because of the badly merged changes from master, I guess I just removed and npm installed everything again) and it screwed things up completely.

I compared the changes from this branch with master and tried to install everything/run the tests locally - it was fine when I noticed that the only difference is lock file and there's something off with it.

So yeah, I thought I am going crazy 🤯🤯🤯

But finally all green.

B4nan:
The thing with reinstalling I noticed some time ago - if you want to wipe node modules, lock file and reinstall, you also need to wipe the node modules folder inside `packages/*`. Dont ask me why, you know my stand on how dumb NPM is :D Just the fact that the yesterdays weirdness was resolved by doing `yarn` instead and checking its error logs is self explanatory :D

packages/actor-scraper/playwright-scraper/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM apify/actor-node:16 AS builder
```
B4nan, line 1: we had a discussion about the dockerfiles yesterday with @mnmkng, we should generally use the same image for the builder, as it will have the browser already preinstalled (and we will pull that image later on anyway)

there might be more improvements, but lets do at least this one thing, we can port the rest of the changes later

B4nan (commented):
gave it just a very bried look, i'd say lets merge and push to the platform as beta to test it properly there

packages/actor-scraper/playwright-scraper/Dockerfile
AndreyBykov: maybe instead of having multistage in this case - we could just remove src folder? What's the point of double installation? Well - dev dependencies will stay, but is it that bad really?

AndreyBykov:
As for the previous comment (node modules wiping etc) - that would actually explain it - I was kinda thinking even how I ended up with that new lock file - as when I tried to wipe it and npm install fresh - it was still different - so yeah - must have left some node_modules in packages

packages/actor-scraper/playwright-scraper/Dockerfile
AndreyBykov: Also, wasn't there some stuff about the access right for node vs browser images? Or should it just work?

AndreyBykov: @B4nan  So I tried to build it from local package files (under my acc, when I was checking whether it's working at all), and it failed, to make sure I tried to build the template actor using the same docker for both stages. Did I get your comment wrong or it's expected?

B4nan: Dev deps are the point, and yes, it is bad, or better say unnecessary. Multistage builds are the correct solution.

But you are right that this might need more changes because of the different user in browser images. We were discussion that too, IMHO we should not change user in base images, that belongs to the project's dockerfile, not to those it extends... 

I guess lets revert it now and get back to it after we have a proper fix elsewhere we can just port in here.

PR #28, opened 2022-08-09
docs: update website design based on crawlee.dev

B4nan:
Redesign of the apify sdk docs based on crawlee.dev + recoloring + quick port of some parts of the content (some guides, examples).

Also contains v1 and v2 docs snapshots with working redirects, so ready to be placed at https://sdk.apify.com.


docs/examples/crawl_multiple_urls_puppeteer.ts
```diff
@@ -8,11 +11,11 @@ const crawler = new PuppeteerCrawler({
...
+// Run the crawler
+await crawler.run([
     'http://www.example.com/page-1',
     'http://www.example.com/page-2',
     'http://www.example.com/page-3',
 ]);
 
-// Run the crawler
-await crawler.run();
+await Actor.init();
```
mnmkng, line 21:
```suggestion
await Actor.exit();
```

docs/examples/crawl_relative_links_all.ts
```diff
@@ -1,4 +1,7 @@
-import { CheerioCrawler, EnqueueStrategy } from '@crawlee/cheerio';
+import { CheerioCrawler, EnqueueStrategy } from 'crawlee';
```
mnmkng, line 1: I find the `EnqueueStrategy` imports a bit too Java-eque. Let's skip them and use just the strings. It's common in Puppeteer and Playwright as well.

PR #32, opened 2022-08-15
chore: login to testing account for e2e tests

vladfrangu:
Will run it first to ensure it works and merge if it passes

EDIT: Marked as draft as we need to discuss what to do about the tests we would run and also need to do some more configuration for them

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -39,6 +39,9 @@ jobs:
                     restore-keys: |
                         turbo-${{ github.job }}-${{ github.ref_name }}-
 
+            -   name: Login to Apify
+                run: npx -y apify-cli login -t ${{ secrets.APIFY_TOKEN }}
+
             -   name: Install Dependencies
                 run: npm ci
 


B4nan:
> EDIT: Marked as draft as we need to discuss what to do about the tests we would run and also need to do some more configuration for them

Discuss what? :]

vladfrangu:
> > EDIT: Marked as draft as we need to discuss what to do about the tests we would run and also need to do some more configuration for them
> 
> Discuss what? :]

Well currently the e2e tests in this repository are just copies of the old ones from crawlee, which feels redundant...

B4nan:
The ones in here are scraper tests, they test the scrapers, not the crawlee classes. We test them via inputs.

vladfrangu:
Ohh, rightt. I will say I'm concerned at the fact the e2e test has been running for 3 hours without anything yielded tho...

B4nan:
Yeah its broken, it needs to finish within an hour id say. It wasnt finishing locally either.

vladfrangu:
Should I adapt the e2e tests to the same system we use on crawlee so we test on platform?

B4nan:
Well, how? We need to test scrapers, not push new actors like we do in crawlee. I guess we could use the client somehow, but we dont want the same setup as with crawlee.

B4nan:
I will merge this as it at least helps with one issue, but the E2E tests are not working correctly at all (the run script).

PR #33, opened 2022-08-15
docs: update homepage

mnmkng:
CleanShot 2022-08-15 at 18 33 05@2x


B4nan (approved):
As said on slack i would make the crawlee logo bigger, otherwise lgtm

B4nan:
We also need to set up the redirect for old getting started guide before we move this to sdk.apify.com

PR #43, opened 2022-08-22
chore(deps): bump puppeteer support to 17.0

vladfrangu:
~~Needs https://github.com/apify/crawlee/pull/1469 first~~

Needs https://github.com/apify/crawlee/pull/1519


B4nan:
we should merge this before shipping new version. i dont think the CI will get fixed itself somehow :]

vladfrangu:
CI won't fix unless we release crawlee first @B4nan so yeah, maybe we should just merge it now

B4nan:
Oh I see, I dont want to breake the build. We can use `npm i -f` if its just about peer deps? Or any other issues?

I would still like to explore conditional builds so we can depend on beta crawlee for dev versions and have the stable there for stable releases only. I feel like we will be falling into this regularly.

vladfrangu:
Hmm, the CI is currently failing because the lock file is out of sync, let me fix that

PR #49, opened 2022-08-24
fix: use correct event manager for actor methods

vladfrangu:
This fixes the actor side of https://github.com/apify/crawlee/issues/1463

For the crawlee only side, I'll open up a PR there

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -246,14 +246,14 @@ export class Actor<Data extends Dictionary = Dictionary> {
      * @ignore
      */
     on(event: EventTypeName, listener: (...args: any[]) => any): void {
-        this.eventManager.on(event, listener);
+        this.config.getEventManager().on(event, listener);
     }
 
     /**
      * @ignore
      */
     off(event: EventTypeName, listener?: (...args: any[]) => any): void {
-        this.eventManager.off(event, listener);
+        this.config.getEventManager().off(event, listener);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


packages/apify/src/actor.ts
```diff
@@ -179,6 +177,9 @@ export class Actor<Data extends Dictionary = Dictionary> {
             this.config.useStorageClient(options.storage);
         }
 
+        // Init the event manager the config uses
```
B4nan, line 180: this looks wrong, we need to use the `PlatformEventManager`, without it the platform events (from websockets) wont work at all.

vladfrangu: The if check above will make the config use the PlatformEventManager, no?

B4nan: Right, I am blind. But then I dont understand why did you have to change it to use the one from config?

B4nan: Or is the fix just about moving the `init` call and the rest is unrelated? i mean the changes from `this.eventManager` to `this.config.getEventManager()`

B4nan: Oh wait a sec, now I get it. We dont want to use the platform implementation locally, but we did.

vladfrangu: Yep, thats the reason!

B4nan (approved):

B4nan (approved):

PR #70, opened 2022-09-14
feat: update the scrapers

AndreyBykov:
- add globs
- update READMEs
- update packages

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16981,17 +16981,15 @@
                 "content-type": "^1.0.4"
             },
             "devDependencies": {
+                "@crawlee/browser-pool": "^3.0.0",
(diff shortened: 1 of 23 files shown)


AndreyBykov:
@B4nan Please check at least briefly puppeteer/web scraper readmes and other stuff. I'm still quite frustrated, so might have missed something (but probably not).

In any case - after it's merged - will rebuild all the scrapers and it's done.

AndreyBykov:
Also note that I updated deps (so that it's aligned between the packages). I set crawlee to ^3.0.0 everywhere, otherwise we should probably set 3.0.4 everywhere (not only in `apify` package).

PR #73, opened 2022-09-20
chore: fix web scraper

AndreyBykov:
- pin puppeteer dependency to version 14
- add `@types/content-type` to deps
- add `"lib": ["dom"]` to ts-config


AndreyBykov:
@B4nan for visibility. I would not dive deep into this, but would rather check puppeteer 17 once new Crawlee/Apify SDK is out.

PR #81, opened 2022-10-04
feat: [WIP] initial add of apify-extra package with dataset functions

metalwarrior665:
Build not working yet

### DIFF
--- a/packages/apify-extra/package.json
+++ b/packages/apify-extra/package.json
@@ -60,5 +60,8 @@
     "peerDependencies": {
         "apify": ">= 3.0.0",
         "crawlee": ">= 3.0.0"
+    },
+    "devDependencies": {
+        "apify": "^3.0.0"
     }
 }
--- a/packages/apify-extra/package.json
+++ b/packages/apify-extra/package.json
@@ -1,0 +1,64 @@
+{
+    "name": "apify-extra",
+    "version": "0.0.1",
+    "description": "Advanced and experimental functionality for Apify Actors and Crawlee. Smaller test coverage. Use with caution!",
+    "engines": {
+        "node": ">=16.0.0"
+    },
+    "main": "./dist/index.js",
+    "module": "./dist/index.mjs",
+    "types": "./dist/index.d.ts",
+    "exports": {
+        ".": {
+            "import": "./dist/index.mjs",
+            "require": "./dist/index.js",
+            "types": "./dist/index.d.ts"
+        },
+        "./package.json": "./package.json"
+    },
+    "keywords": [
+        "apify",
+        "headless",
+        "chrome",
+        "puppeteer",
+        "crawler",
+        "scraper"
+    ],
+    "author": {
+        "name": "Apify",
+        "email": "support@apify.com",
+        "url": "https://apify.com"
+    },
+    "contributors": [
+        "Jan Curn <jan@apify.com>",
+        "Marek Trunkat <marek@apify.com>",
+        "Ondra Urban <ondra@apify.com>"
+    ],
+    "license": "Apache-2.0",
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/apify/apify-sdk-js"
+    },
+    "bugs": {
+        "url": "https://github.com/apify/apify-sdk-js/issues"
+    },
+    "homepage": "https://sdk.apify.com",
+    "scripts": {
+        "build": "npm run clean && npm run compile && npm run copy && npm run fixApifyExport",
+        "clean": "rimraf ./dist",
+        "compile": "tsc -p tsconfig.build.json && gen-esm-wrapper ./dist/index.js ./dist/index.mjs",
+        "copy": "ts-node -T ../../scripts/copy.ts --readme=local",
+        "fixApifyExport": "ts-node -T ../../scripts/temp_fix_apify_exports.ts"
+    },
+    "publishConfig": {
+        "access": "public"
+    },
+    "dependencies": {
+        "@types/bluebird": "^3.5.37",
+        "bluebird": "^3.7.2"
+    },
+    "peerDependencies": {
+        "apify": ">= 3.0.0",
+        "crawlee": ">= 3.0.0"
+    }
+}
--- a/packages/apify-extra/package.json
+++ b/packages/apify-extra/package.json
@@ -53,10 +53,6 @@
     "publishConfig": {
         "access": "public"
     },
-    "dependencies": {
-        "@types/bluebird": "^3.5.37",
-        "bluebird": "^3.7.2"
-    },
     "peerDependencies": {
         "apify": ">= 3.0.0",
         "crawlee": ">= 3.0.0"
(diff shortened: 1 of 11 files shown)


packages/apify-extra/package.json
```diff
@@ -0,0 +1,64 @@
...
+        "compile": "tsc -p tsconfig.build.json && gen-esm-wrapper ./dist/index.js ./dist/index.mjs",
+        "copy": "ts-node -T ../../scripts/copy.ts --readme=local",
+        "fixApifyExport": "ts-node -T ../../scripts/temp_fix_apify_exports.ts"
+    },
+    "publishConfig": {
+        "access": "public"
+    },
+    "dependencies": {
+        "@types/bluebird": "^3.5.37",
+        "bluebird": "^3.7.2"
```
B4nan, line 58: do we really need to use bluebird 😭

packages/apify-extra/src/dataset.ts
```diff
@@ -0,0 +1,339 @@
...
+        await Actor.setValue(PROCESS_FN_LOADING_STATE_KV_RECORD_KEY, processFnLoadingState);
+    });
+
+    const requestInfoArr = await createRequestArray(processFnLoadingState);
+    if (debugLog) {
+        log.info(`Number of requests to do: ${requestInfoArr.length}`);
+    }
+
+    //  Now we execute all the requests in parallel (with defined concurrency)
+    await bluebird.map(requestInfoArr, async (requestInfoObj) => {
```
B4nan, line 264: should be the same as:

```suggestion
    await Promise.all(requestInfoArr.map(async (requestInfoObj) => {
```

and we can let bluebird rest in its graveyard

packages/apify-extra/src/dataset.ts
```diff
@@ -0,0 +1,339 @@
+import { Actor } from 'apify';
+import { Dictionary, log, sleep } from 'crawlee';
+import bluebird from 'bluebird';
+import type { Dataset } from 'apify-client';
```
B4nan, line 4: undeclared dependency (as long as this import is preserved in the d.ts file, we need to declare it in regular deps, not just dev deps)

but maybe better to get rid of this type instead of declaring it, i am sure we have some equivalent in the `@crawlee/types`

B4nan (changes requested):
you will need to run `npm i` in the root to update the lock file and push that

image

packages/apify-extra/package.json
metalwarrior665: I will eventually get rid of it but it will require a bit of work

packages/apify-extra/src/dataset.ts
metalwarrior665: Nope, it is there for the concurrency option. I can replace it later with AutoscaledPool, BasicCrawler or maybe I will find something more lightweight

barjin:
Ok, after partially revamping this yesterday, I have a few points/questions (@metalwarrior665 ):
- Migration-safe dataset upload - what is the use case? The local "context" wipes with every migration nevertheless. This question is just out of curiosity though, I can see this working.
- `loadDatasetItemsInParallel` - if I understand correctly, this:
  - a) loads all the items from the specified datasets, optionally combines those and returns the items
    - Use case? We have to wait till the loading ends either way - and when the migration happens somewhere in between, we can start anew. 
  - b) if provided with a `processFn` function, it runs this function on all of the items -> but then doesn't return anything - in this case, it keeps track of the processed datasets/chunks in case of a migration (this makes sense!)

If I can add my two cents, I like the "parallel" forEach (map?) approach - it reminds me of Hadoop MapReduce and similar Big Data processing systems - and imo works well with an environment with migrations; for the rest, I am ready to have my mind changed :)

packages/apify-extra/package.json
```diff
@@ -60,5 +60,8 @@
     "peerDependencies": {
         "apify": ">= 3.0.0",
         "crawlee": ">= 3.0.0"
+    },
+    "devDependencies": {
+        "apify": "^3.0.0"
```
B4nan, line 65: i doubt its about this, the package is locally defined and mapped via tsconfig, this shouldnt be really used locally as paths mapping should have preference.

barjin: imo it's a "race condition" coming from `turbo`  - without the declared dependency, `turbo` builds `apify` and `apify-extra` in parallel (`rimraf`? I'm a bit puzzled why would it look into the `dist` folder, but I should just read up on this more).

Anyway, if you prebuild `apify` with 
`npx turbo run build --filter=apify `
and then go for `npm run build` (with the `apify` build cached), the build finishes w/o problems.

B4nan: thats not really a solution, workaround at best. turbo should respect peer dependencies and resolve things correctly, thats what its for. if you think its caused by turbo doing things in parallel, try disabling the concurrency, but i doubt it.

barjin: It is caused by turborepo: 
https://github.com/vercel/turborepo/issues/1601

Rebased this branch here, specifying this dependency in `turbo.json`:
https://github.com/apify/apify-sdk-js/pull/108

B4nan: i can see it fails the same with lerna without the explicit dev dep, weird but tolerable. weird, iirc i was trying this out already. there are other build failures but those are probably caused by the PR not being rebased

B4nan: lets use the dev dependency, better than adjusting turbo config i guess, i thought it wasnt enough

packages/apify-extra/src/dataset.ts
barjin: [something more lightweight?](https://github.com/apify/apify-sdk-js/pull/81/commits/f156a7f67357b53aea4b1da9b84cccf93d475f42#diff-f2f82b79930ea9b421746fa98f15f05cdd0c98f879ed1116dd1dfafb69740783R9)

metalwarrior665:
@barjin 
1. Use-case is when you need to push a lot of items at once. Typical use-case is dataset transformations, you load 1M items and push 1M so you need to increase speed via paralelism and make sure you only start where you ended (loading is much faster than pushing)
2.
   a) Use-case is that it keeps the returned items in order (in datasets and inside dataset) and the parallel loading and that it is able to parallelize over more dataset and also inside one dataset. You can go even 30x speed this way. If migration happens, you have to load again yeah but with the speed you have a good chance it doesn't happen.
   b) This part was a bit messy so will need good tests

We could get rid of a) for simplicity and make `processFn` mandatory but then the keeping track of migration needs to be optional because sometimes you want to start from scratch. And if the user want the items in order, they would need to do that themselves by checking which dataset the chunk is from and what are the offsets. So since we already provide that functionality (if you transform dataset it might be nice that it is in the same order) I would keep it but maybe we can change the interface/implementation. The function definitely ridiculously complex and needs to be refactored eventually and tested in smaller pieces.

metalwarrior665:
My plan going forward is to have one more look on the features and the logic (suggestions welcome) and then we would release it in beta so we can start using it internally. Then add proper tests before we release it in latest. Sounds good?

metalwarrior665:
Maybe the way to do this better would be to extend the SDK classes so it is more in line with SDK rather than having standalone functions with tons of options. But I probably don't have capacity for a major rewrite now :)

barjin:
closed in favor of #116, all further development there

PR #83, opened 2022-10-07
feat(apify): add decryption for input secrets

drobnikj:
Implementation for decryption of input secrets.

- [x] move APIFY_INPUT_SECRETS_PRIVATE_KEY_PASSPHRASE and APIFY_INPUT_SECRETS_PRIVATE_KEY_FILE into @apify/consts

I still didn't comment or document the secrets at all, I would do it once we test it internally in team and release it into public.


packages/apify/src/actor.ts
```diff
@@ -646,7 +648,16 @@ export class Actor<Data extends Dictionary = Dictionary> {
      * @ignore
      */
     async getInput<T = Dictionary | string | Buffer>(): Promise<T | null> {
-        return this.getValue<T>(this.config.get('inputKey'));
+        const inputSecretsPrivateKeyFile = this.config.get('inputSecretsPrivateKeyFile');
+        const input = await this.getValue<T>(this.config.get('inputKey'));
+        if (inputSecretsPrivateKeyFile && ow.isValid(input, ow.object.nonEmpty) && inputSecretsPrivateKeyFile) {
```
fnesveda, line 653: `inputSecretsPrivateKeyFile` is here twice

fnesveda: But maybe you should check here for `inputSecretsPrivateKeyPassphrase` instead, it could happen that someone would pass only the secrets file but not the passphrase, and then it would break.

fnesveda (changes requested):
Looks good! I found just one small thing.

package.json
```diff
@@ -64,8 +66,6 @@
         "@typescript-eslint/parser": "5.39.0",
         "commitlint": "^17.0.3",
         "crawlee": "^3.0.0",
-        "playwright": "*",
-        "puppeteer": "*",
```
fnesveda, line 68: I think these should stay here, no? They're used in the e2e tests.

fnesveda (changes requested):

package.json
fnesveda: I think it's gonna work, since they're still in the lockfile, but they're imported from the scraper source code, which is then imported dynamically through here: https://github.com/apify/apify-sdk-js/blob/6d1af7916e576e30abd9a494881540f1598447a2/test/e2e/tools.mjs#L58

But I don't know how it works with the workspaces, if you reference a workspace which has them listed as dependencies, maybe you don't need to list them as your own dependencies? I guess it's fine in that case then.

drobnikj: Test passed so it should be fine.

fnesveda (approved):

packages/apify/src/configuration.ts
```diff
@@ -18,6 +18,9 @@ export interface ConfigurationOptions extends CoreConfigurationOptions {
     proxyStatusUrl?: string;
     isAtHome?: boolean;
     userId?: string;
+    inputSecretsPrivateKeyPassphrase?: string;
+    inputSecretsPrivateKeyFile?: string;
+    inputKey?: string;
```
B4nan, line 23: not sure if it makes sense, but this option is apparently not mapped from env vars

test/apify/actor.test.ts
```diff
@@ -1,5 +1,7 @@
+import { createPublicKey } from 'crypto';
```
B4nan, line 1:
```suggestion
import { createPublicKey } from 'node:crypto';
```

B4nan (approved):

packages/apify/src/configuration.ts
drobnikj: It is probably not necessary here as the key is in crawlee core CoreConfigurationOptions.

Issue #84, opened 2022-10-07
Flaky implementation some of Actor tests


drobnikj:
### Issue description

Some of the tests for Actor didn't work correctly, the problem is in mocking getValue method on top of defaultStore. The expect calls are in mock function, but the mock function is never called so the test passed even if there is a bug, see https://github.com/apify/apify-sdk-js/blob/master/test/apify/actor.test.ts#L983 
Plus the test for `Actor.pushData` testing setValue method.

I just discover it during #83, but don't have time to fix it.

### Package version

3.0.2

### Node.js version

all

B4nan:
> The expect calls are in mock function,

This is so common problem, it cought me several times too. Would be great to have some eslint rule for assertions inside async callbacks.

@vladfrangu ptal

PR #95, opened 2022-10-12
chore: convert some tests to use spies instead of property overrides

vladfrangu:
Tackles a part of #84


test/apify/actor.test.ts
```diff
@@ -991,44 +991,36 @@ describe('Actor', () => {
...
+            const setValueSpy = jest.spyOn(defaultStore, 'setValue');
 
-            const oldSet = defaultStore.setValue;
-            defaultStore.setValue = async (key, value) => {
-                expect(key).toBe('key-1');
-                expect(value).toBe(record);
-            };
+            setValueSpy.mockImplementation(async (key, value) => {
+                expect(key).toEqual('key-1');
+                expect(value).toEqual(record);
```
B4nan, line 998: those assertions will result in promise rejection instead of test failure, that's another part of the problem in that issue

B4nan (approved):

PR #107, opened 2022-10-19
chore: update generic scrapers readme and input schema

AndreyBykov:
- Remove Pseudo-Urls from readme/input
- add v3 breaking changes (enforced proxy usage, deprecation of pseudo-urls)

### DIFF
--- a/packages/actor-scraper/cheerio-scraper/INPUT_SCHEMA.json
+++ b/packages/actor-scraper/cheerio-scraper/INPUT_SCHEMA.json
@@ -1,7 +1,7 @@
 {
     "title": "Cheerio Scraper Input",
     "type": "object",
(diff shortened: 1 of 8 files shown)


B4nan (commented):
let me ask for final confirmation both @mnmkng and @jancurn, I remember there was some push back when we initially wanted to remove it (from the library), so just wanted to double check its fine to remove it from the input schema. 

also, to me this feels kind of weird, to remove it from the schema and call it deprecated - deprecated things can be still used, but here we are basically removing the UI, hence it cannot be used (by regular users)

AndreyBykov:
Yeah, I also thought about those things. We could either remove that it's deprecated and keep as legit option, but I guess we really want to get rid of it. If that's the case - removing it from inputs keeps it working for existing scrapers, but new user would not have a choice other than using globs. But yeah - I just created this PR as it was mentioned in Slack to remove it.

jancurn:
Guys, feel free to deprecate PseudoURLs but don't remove it yet, because you will piss off a ton of our core users. Keep in mind that these are our most grossing public actors; they really are our core product, and removing features people depend, without any advance warning or analysis of how many people depend on them, will break the trust of our users and drive them away.

B4nan:
Yeah, that's what I thought, I'd keep it as is (if we want to remove the support at some point, e.g. in v4), or remove the "(deprecated)" part if we plan on keeping it for longer (no harm in keeping it deprecated for more than one major version).

AndreyBykov:
Will do another small PR with removed `(deprecated)` and added `changes` 👍

PR #113, opened 2022-10-24
chore(web-scraper): store more data in default page function

zpelechova:
Added h1, first h2 and random text from the page to show better what web scraper can do.
Internal communication here: https://apifier.slack.com/archives/C0319CUGRA7/p1666609016592329?thread_ts=1666258262.608509&cid=C0319CUGRA7


B4nan (approved):
Maybe worth adjusting pw/pup/cheerio scrapers too?

PR #116, opened 2022-10-28
feat: [WIP] apify extra

barjin:


metalwarrior665:
@barjin Are we still stuck on something or there wasn't the capacity to finalize? I will definitely want to polish the code once more but also would like to do it when we know how to build it :)

barjin:
feel free to review, guess it sort of fizzled out @B4nan ? 

btw it's been buildable since the first day, it's all kinda finished now - correctness and usability remain to be checked by someone with more actor-building expertise :)

metalwarrior665:
Hah, I didn't even realize you refactored the whole thing and added tests :) Let me just review it and we can release, good job!

metalwarrior665:
btw @barjin @B4nan I'm thinking about this one from time to time. Last year, there were only few new projects so there wasn't much need to implement any advanced functionality (and what we needed was fine to copy/paste).

In the end, we can probably just keep copy-pasting for now. If we want to push this through, I think it would be better to do it right into Crawlee, the interface to Dataset is the same and it will bubble into SDK.

The only other one that is used quite frequently is parallelPersistedCall ([see example here](https://github.com/apify-projects/BeeDigital/blob/master/search-terms-master-runner/src/persisted-parallel-call.ts)) which should probably go to the SDK. The client would be cool but it doesn't understand migration which is needed in this case.

barjin:
Closing as this got (really) stale

PR #122, opened 2022-11-02
chore: log crawlee version in system info

vladfrangu:


packages/apify/src/utils.ts
```diff
@@ -6,6 +6,8 @@ import semver from 'semver';
 // @ts-expect-error if we enable resolveJsonModule, we end up with `src` folder in `dist`
 import { version as apifyClientVersion } from 'apify-client/package.json';
 // @ts-expect-error if we enable resolveJsonModule, we end up with `src` folder in `dist`
+import { version as crawleeVersion } from 'crawlee/package.json';
```
B4nan, line 9: we should check it from the core package i guess, the meta package might not be there and this would fail

vladfrangu: Hrm, good point. We'll need a release of /core that exports the package.json first (I'll open that rn) then we can do this!

PR #152, opened 2022-12-06
fix: jsdomscraper runScripts option + log suppressing

barjin:
Adds the `runScripts` option and the option to turn on the internal JSDOM console logs (off by default). 

Depends on [@apify/crawlee#1707](https://github.com/apify/crawlee/pull/1707).


packages/actor-scraper/jsdom-scraper/src/internals/crawler_setup.ts
```diff
@@ -158,6 +158,8 @@ export class CrawlerSetup implements CrawlerSetupOptions {
             proxyConfiguration: this.proxyConfiguration,
             requestHandler: this._requestHandler.bind(this),
             preNavigationHooks: [],
+            runScripts: this.input.runScripts ?? false,
+            hideInternalConsole: !this.input.showInternalConsole ?? true,
```
B4nan, line 162: the defaults in the input schema are saying the opposite, so i guess it should be like this:
```suggestion
            runScripts: this.input.runScripts ?? true,
            hideInternalConsole: !this.input.showInternalConsole ?? false,
```

B4nan (changes requested):

B4nan:
also needs a manual lock file update i guess

packages/actor-scraper/jsdom-scraper/src/internals/crawler_setup.ts
barjin: Huh, I swear the `!` vs `??` precedence seemed better in my head :)

PR #169, opened 2023-02-06
feat(Puppeteer/PlaywrightScraper): add support for headful crawling

vladfrangu:
Handles a long lasting request. I would love if you had any better descriptions for the option, as I don't fully know how to describe it past...just linking to an article explaining it

### DIFF
--- a/packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
+++ b/packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
@@ -118,11 +118,11 @@
             "groupCaption": "Browser masking",
             "groupDescription": "Settings that help mask as a real user and prevent scraper detection."
         },
-        "headful": {
-            "title": "Run browsers in headful mode",
+        "headless": {
+            "title": "Run browsers in headless mode",
             "type": "boolean",
-            "description": "By default, browsers run in headless mode. You can toggle this on to run them in headful mode. Headful mode can help with certain rare anti-scraping protections but is slower and more costly.",
-            "default": false
+            "description": "By default, browsers run in headless mode. You can toggle this off to run them in headful mode, which can help with certain rare anti-scraping protections but is slower and more costly.",
+            "default": true
         },
         "ignoreSslErrors": {
             "title": "Ignore SSL errors",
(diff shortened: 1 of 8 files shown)


packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
```diff
@@ -118,6 +118,12 @@
             "groupCaption": "Browser masking",
             "groupDescription": "Settings that help mask as a real user and prevent scraper detection."
         },
+        "headful": {
+            "title": "Run browsers in headful mode",
+            "type": "boolean",
+            "description": "By default, browsers run in headless mode. You can toggle this on to run them in headful mode.",
```
metalwarrior665, line 124: I would just add a sentence as `Headful mode can help with certain rare anti-scraping protections but is slower and more costly.`

metalwarrior665 (changes requested):
Just a small nitpick

AndreyBykov:
I would maybe instead of naming it `headful` and disabling by default name it `headless` and enable it by default, as term `headful` is rather niche. It's used more now, but from my experience it's usually `headless` or `full` in discussions/docs, and e.g. playwright does not have `headful` in docs at all. But that's rather a nit.

vladfrangu:
That makes more sense, you're right. I'll apply it :D

AndreyBykov:
@vladfrangu let's see what others think - as we use it internally from the beginning of time =D

vladfrangu:
> @vladfrangu let's see what others think - as we use it internally from the beginning of time =D

Welp, already pushed a commit 😆 If others agree with keeping it headless then yay, if not, just undo one commit and we're done :D

metalwarrior665:
I would vote slightly more for headful as it is the thing you opt-in (and you don't have to think in double-negation) but not a big deal :)

B4nan (approved):

vladfrangu:
Maybe we can make a poll or something internally about this? What do you think?

metalwarrior665:
@vladfrangu It is not that important I think :) Just go with your gut :)

vladfrangu:
I'm going to be merging this as is later today - to me it feels more accurate to name the option what it is (headless), defaulted to true, and it is also how users reading playwright/puppeteer docs would see it, but if you disagree let me know :D

PR #172, opened 2023-02-14
feat: terminal message on Actor.exit()

barjin:
Depends on [@apify/apify-client-js#306](https://github.com/apify/apify-client-js/pull/306).


packages/apify/src/actor.ts
```diff
@@ -512,15 +514,15 @@ export class Actor<Data extends Dictionary = Dictionary> {
      * For more information, see the [Actor Runs](https://docs.apify.com/api/v2#/reference/actor-runs/) API endpoints.
      * @ignore
      */
-    async setStatusMessage(statusMessage: string): Promise<ClientActorRun> {
+    async setStatusMessage(statusMessage: string, isStatusMessageTerminal?: boolean): Promise<ClientActorRun> {
```
B4nan, line 517: let's do options object

B4nan (approved):

barjin:
The last commit contains a "breaking" change in types - if somebody uses helper types to extract the return types of our methods. ideas welcome

packages/apify/src/actor.ts
```diff
@@ -512,15 +514,21 @@ export class Actor<Data extends Dictionary = Dictionary> {
      * For more information, see the [Actor Runs](https://docs.apify.com/api/v2#/reference/actor-runs/) API endpoints.
      * @ignore
      */
-    async setStatusMessage(statusMessage: string): Promise<ClientActorRun> {
+    async setStatusMessage(statusMessage: string, options?: { isStatusMessageTerminal?: boolean }): Promise<any> {
```
B4nan, line 517: we should revert this bit, the return type was correct here
```suggestion
    async setStatusMessage(statusMessage: string, options?: { isStatusMessageTerminal?: boolean }): Promise<ClientActorRun> {
```

packages/apify/src/actor.ts
```diff
@@ -1138,7 +1146,7 @@ export class Actor<Data extends Dictionary = Dictionary> {
      * @returns The return value is the Run object.
      * For more information, see the [Actor Runs](https://docs.apify.com/api/v2#/reference/actor-runs/) API endpoints.
      */
-    static async setStatusMessage(statusMessage: string): Promise<ClientActorRun> {
+    static async setStatusMessage(statusMessage: string): Promise<any> {
```
B4nan, line 1149:
```suggestion
    static async setStatusMessage(statusMessage: string): Promise<ClientActorRun> {
```

packages/apify/src/actor.ts
```diff
@@ -512,15 +514,21 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+    async setStatusMessage(statusMessage: string, options?: { isStatusMessageTerminal?: boolean }): Promise<ClientActorRun> {
+        const { isStatusMessageTerminal } = options || {};
         ow(statusMessage, ow.string);
+        ow(isStatusMessageTerminal, ow.optional.boolean);
 
         const runId = this.config.get('actorRunId')!;
         if (!runId) {
+            if (!this.isAtHome()) {
+                log.info(`Setting status message to ${statusMessage}...`);
+                return;
```
B4nan, line 526: this might need some polishment, either return some empty object like response, or ideally rework it so it can pass the call to current storage implementation

packages/apify/package.json
```diff
@@ -62,10 +62,11 @@
         "@crawlee/core": "^3.1.1",
         "@crawlee/types": "^3.1.1",
         "@crawlee/utils": "^3.1.1",
-        "apify-client": "^2.6.0",
+        "apify-client": "^2.6.3",
+        "crawlee": "^3.3.0",
```
B4nan, line 66: i guess this is here mistakenly, right?

packages/apify/src/actor.ts
```diff
@@ -512,15 +514,40 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
-            throw new Error(`Environment variable ${ENV_VARS.ACTOR_RUN_ID} is not set!`);
+        if (runId) {
+            const run = await this.apifyClient.run(runId).get();
+            if (run) {
+                return run;
+            }
         }
 
-        return this.apifyClient.run(runId).update({ statusMessage });
+        return null as unknown as ClientActorRun;
```
B4nan, line 550: maybe empty object will be better than `null`? at least it wont fail on property access

PR #179, opened 2023-03-23
fix: log status message only once and without prefix

barjin:
Removes the `[Status message]` prefix from the actual status message, leaves it in the logs.


barjin:
Yea, I got a bit lost in Lukas' repro, but this should be it.

B4nan (approved):

PR #187, opened 2023-04-24
feat(cheerio): expose transformRequestFunction as hidden input

metalwarrior665:


mtrunkat (approved):

metalwarrior665:
I would not merge it as it is since combining it with the default behavior is hard

B4nan:
Maybe we should add this to all the scrapers?

PR #188, opened 2023-04-27
Feat: cheerio experimental queue

petrpatek:
This is a POC version with minimal API and minimal request footprint.

Possible improvements: 

- Use magazine instead of the queue to allow syncing with kv-store during the run and save only the rest loaded in memory. 
- compress the requested objects in memory
- use `large-set` package to bypass nodeJs internal limit for Set


metalwarrior665:
Let's keep this non-merged for some time.

packages/actor-scraper/cheerio-scraper/src/internals/request_queue_experimental.ts
```diff
@@ -0,0 +1,186 @@
...
+            .sort(({ key: key1 }, { key: key2 }) => this._sortByChunkOrder(key1, key2));
+
+        let finalItems: any[] = [];
+        for (const { key } of orderedKeys) {
+            const items = await Actor.getValue(key);
+            finalItems = finalItems.concat(items);
+        }
+
+        await Promise.all(orderedKeys.map(({ key }) => {
+            return storeClient.deleteRecord(key);
```
metalwarrior665, line 145: This could be dangerous if you crash right after this delete, you loose everything. I would prefer to figure out how to overwrite properly.

packages/actor-scraper/cheerio-scraper/src/internals/queue.ts
```diff
@@ -0,0 +1,27 @@
+export default class Queue<T> {
```
metalwarrior665, line 1: Add a comment explaining where is this used, wasn't obvious at first

metalwarrior665 (changes requested):
Looks great. Should have a few more comments. Nothing urgent with the remarks.

PR #203, opened 2023-06-08
fix: only print status message when explicitly provided in `Actor.exit()`

B4nan:
The default (generic) status message from `Actor.exit` was always overriding the final status message from basic crawler (that has much more information). This removes the default messages from there, keeping the one from basic crawler.

Sidenote: maybe we should also add current concurrency to the status messages?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,5 +2,4 @@
 # See: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
 
 # To keep eye on new functions added to Actor specification
-/packages/apify/src/            @jancurn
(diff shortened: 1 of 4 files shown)


B4nan:
Hmm, apparently, the codeowners thing won't add multiple people (even in a public repo).

B4nan:
> Hmm, apparently, the codeowners thing won't add multiple people (even in a public repo).

So it should be a single line I guess.

https://stackoverflow.com/questions/66821568/require-github-pr-approvals-from-multiple-groups

metalwarrior665 (approved):

mnmkng (approved):

PR #209, opened 2023-06-20
refactor: port WCC features to generic (PW/PP/Web) scraper actors

### LINKED ISSUE #208: Add new options to generic scrapers
### Which scraper is the feature request for?

cheerio-scraper

### Feature

We have new features in website content crawler that are worth porting to the generic scrapers, namely:
- exclude globs
- cookie modal closing
- infinite scroll

### Motivation

.

.

_No response_

### Other context

_No response_
```[tasklist]
### Tasks
```

barjin:
Closes #208 

Probably needs some second thought on the cookie closing - the WCC actor has a thought-through waiting logic, which we have to(?) supply with ugly hard waits here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,8 +45,8 @@
         "release": "npm run build && lerna publish from-package --contents dist",
         "publish:next": "lerna publish --canary --preid beta --dist-tag next",
         "release:next": "npm run build && npm run publish:next",
-        "lint": "eslint packages/*/src packages/*/*/src test",
-        "lint:fix": "eslint packages/*/src packages/*/*/src test --fix"
+        "lint": "eslint 'packages/**/src/**/*' test",
+        "lint:fix": "eslint 'packages/**/src/**/*' test --fix"
     },
     "devDependencies": {
         "@apify/consts": "^2.4.1",
--- a/package.json
+++ b/package.json
@@ -45,8 +45,8 @@
         "release": "npm run build && lerna publish from-package --contents dist",
         "publish:next": "lerna publish --canary --preid beta --dist-tag next",
         "release:next": "npm run build && npm run publish:next",
-        "lint": "eslint packages/*/src test",
-        "lint:fix": "eslint packages/*/src test --fix"
+        "lint": "eslint packages/*/src packages/*/*/src test",
+        "lint:fix": "eslint packages/*/src packages/*/*/src test --fix"
     },
     "devDependencies": {
         "@apify/consts": "^2.4.1",
--- a/package.json
+++ b/package.json
@@ -45,8 +45,8 @@
         "release": "npm run build && lerna publish from-package --contents dist",
         "publish:next": "lerna publish --canary --preid beta --dist-tag next",
         "release:next": "npm run build && npm run publish:next",
-        "lint": "eslint 'packages/**/src/**/*' test",
-        "lint:fix": "eslint 'packages/**/src/**/*' test --fix"
+        "lint": "eslint \"packages/**/src/**/*\" test",
+        "lint:fix": "eslint \"packages/**/src/**/*\" test --fix"
     },
     "devDependencies": {
         "@apify/consts": "^2.4.1",
--- a/packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
+++ b/packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
@@ -148,6 +148,12 @@
             "description": "Scraper will ignore CSP (content security policy) and CORS (cross origin resource sharing) settings of visited pages and requested domains. This enables you to freely use XHR/Fetch to make HTTP requests from the scraper.",
             "default": false
         },
+        "closeCookieModals": {
+            "title": "Dismiss cookie modals",
+            "type": "boolean",
+            "description": "Using the [I don't care about cookies](https://addons.mozilla.org/en-US/firefox/addon/i-dont-care-about-cookies/) browser extension. When on, the crawler will automatically try to dismiss cookie consent modals. This can be useful when crawling European websites that show cookie consent modals. This setting is ignored when using the cheerio crawler type.",
+            "default": false
+        },
         "downloadMedia": {
             "sectionCaption": "Performance and limits",
             "title": "Download media",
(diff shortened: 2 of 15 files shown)


packages/actor-scraper/playwright-scraper/INPUT_SCHEMA.json
```diff
@@ -245,6 +245,18 @@
             "prefill": "// We need to return array of (possibly async) functions here.\n// The functions accept a single argument: the \"crawlingContext\" object.\n[\n    async (crawlingContext) => {\n        const { page } = crawlingContext;\n        // ...\n    },\n]",
             "editor": "javascript"
         },
+        "closeCookieModals": {
+            "title": "Dismiss cookie modals",
+            "type": "boolean",
+            "description": "Using the [I don't care about cookies](https://addons.mozilla.org/en-US/firefox/addon/i-dont-care-about-cookies/) browser extension. When on, the crawler will automatically try to dismiss cookie consent modals. This can be useful when crawling European websites that show cookie consent modals. This setting is ignored when using the cheerio crawler type.",
```
B4nan, line 251: this bit is not valid here, right?

```suggestion
            "description": "Using the [I don't care about cookies](https://addons.mozilla.org/en-US/firefox/addon/i-dont-care-about-cookies/) browser extension. When on, the crawler will automatically try to dismiss cookie consent modals. This can be useful when crawling European websites that show cookie consent modals.",
```

packages/actor-scraper/playwright-scraper/src/internals/crawler_setup.ts
```diff
@@ -309,6 +310,16 @@ export class CrawlerSetup implements CrawlerSetupOptions {
         };
         const { context, state } = createContext(contextOptions);
 
+        if(this.input.closeCookieModals) {
```
B4nan, line 313: I am sure this could be caught by eslint, crawlee reports that as `keyword-spacing` error.

```suggestion
        if (this.input.closeCookieModals) {
```

packages/actor-scraper/puppeteer-scraper/INPUT_SCHEMA.json
```diff
@@ -233,6 +233,18 @@
             "prefill": "// We need to return array of (possibly async) functions here.\n// The functions accept a single argument: the \"crawlingContext\" object.\n[\n    async (crawlingContext) => {\n        const { page } = crawlingContext;\n        // ...\n    },\n]",
             "editor": "javascript"
         },
+        "closeCookieModals": {
+            "title": "Dismiss cookie modals",
+            "type": "boolean",
+            "description": "Using the [I don't care about cookies](https://addons.mozilla.org/en-US/firefox/addon/i-dont-care-about-cookies/) browser extension. When on, the crawler will automatically try to dismiss cookie consent modals. This can be useful when crawling European websites that show cookie consent modals. This setting is ignored when using the cheerio crawler type.",
```
B4nan, line 239: same here
```suggestion
            "description": "Using the [I don't care about cookies](https://addons.mozilla.org/en-US/firefox/addon/i-dont-care-about-cookies/) browser extension. When on, the crawler will automatically try to dismiss cookie consent modals. This can be useful when crawling European websites that show cookie consent modals.",
```

B4nan (changes requested):

packages/actor-scraper/playwright-scraper/src/internals/crawler_setup.ts
```diff
@@ -309,6 +310,16 @@ export class CrawlerSetup implements CrawlerSetupOptions {
...
+            await sleep(500);
+            await crawlingContext.page.evaluate(getInjectableScript());
+            await sleep(2000);
```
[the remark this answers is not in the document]
barjin: The problem is that the cookie modal
- has to exist on the page at the moment of injecting the script (loading it dynamically takes some time).
- shouldn't exist on the page when running the user function - while the CSS tricks hide it immediately, the JS clicking scripts take some time. 

Now that I'm thinking about it, there has to be a way of telling whether the script has finished removing the modals. I will look into it, the current implementation is probably not "blocking" and requiring the timeout. 

We can probably merge this now and simply remove the wait once we have the execution blocking in place?

package.json
```diff
@@ -45,8 +45,8 @@
         "release": "npm run build && lerna publish from-package --contents dist",
         "publish:next": "lerna publish --canary --preid beta --dist-tag next",
         "release:next": "npm run build && npm run publish:next",
-        "lint": "eslint packages/*/src test",
-        "lint:fix": "eslint packages/*/src test --fix"
+        "lint": "eslint packages/*/src packages/*/*/src test",
+        "lint:fix": "eslint packages/*/src packages/*/*/src test --fix"
```
B4nan, line 49: FYI the globs should be wrapped in quotes, otherwise it gets expanded by shell which can have quite significant perf cost

https://github.com/apify/crawlee/blob/master/package.json#L49-L50

https://eslint.org/docs/latest/use/command-line-interface#run-the-cli

packages/actor-scraper/playwright-scraper/src/internals/crawler_setup.ts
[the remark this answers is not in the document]
B4nan: Sure, as long as its opt-in, its fine.

package.json
barjin: Ah, nice, eslint can process the globs itself, that simplifies it 👍️

barjin:
Exclude globs should be in all the scrapers (see [commit here](https://github.com/apify/apify-sdk-js/commit/4ee014f05bae11782bf5a14a2daa08e3b0d7d268)), @vladfrangu did that on a user request a month or two ago. Not sure why it's not released in the 3.0 branch though, all the new features are only in `development` (at least for cheerio).

package.json
```diff
@@ -45,8 +45,8 @@
         "release": "npm run build && lerna publish from-package --contents dist",
         "publish:next": "lerna publish --canary --preid beta --dist-tag next",
         "release:next": "npm run build && npm run publish:next",
-        "lint": "eslint packages/*/src test",
-        "lint:fix": "eslint packages/*/src test --fix"
+        "lint": "eslint 'packages/**/src/**/*' test",
+        "lint:fix": "eslint 'packages/**/src/**/*' test --fix"
```
B4nan, line 49: I think we should be using double quotes, the eslint docs suggest that because of windows users too

```suggestion
        "lint": "eslint \"packages/**/src/**/*\" test",
        "lint:fix": "eslint \"packages/**/src/**/*\" test --fix"
```

B4nan (approved):

B4nan:
Will merge it as `refactor` instead of `feat`, as I have a hunch this would be added to the SDK changelog too otherwise.

PR #216, opened 2023-07-19
feat: Use Actor env vars

jirimoravcik:


packages/apify/src/actor.ts
```diff
@@ -836,9 +836,27 @@ export class Actor<Data extends Dictionary = Dictionary> {
     }
 
     /**
-     * Returns a new {@apilink ApifyEnv} object which contains information parsed from all the `APIFY_XXX` environment variables.
+     * Modifies Actor env vars so parsing respects the structure of {@apilink ApifyEnv} interface.
+     */
+    private getModifiedActorEnvVars = () => {
```
B4nan, line 841: i'd keep this as a regular method instead of function expression, to be consistent - we dont use this anywhere (unless really needed)

B4nan (approved):
looking good, just one nit

fnesveda (approved):

packages/apify/src/actor.ts
```diff
@@ -846,9 +846,27 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+            // Prepend `ACTOR_` to env vars so ApifyEnv structure is preserved
+            if (['ID', 'RUN_ID', 'TASK_ID'].includes(k)) {
+                modifiedActorEnvVars[`ACTOR_${k}`] = v;
+            } else {
+                modifiedActorEnvVars[k] = v;
+            }
+        });
+
+        return modifiedActorEnvVars;
+    };
```
B4nan, line 864:
```suggestion
    }
```

PR #217, opened 2023-07-20
docs: Update copy

jancurn:


website/src/pages/index.js
```diff
@@ -20,19 +20,19 @@ function Hero() {
                                 Apify SDK for JavaScript<br /> is a toolkit for<br /> building actors
                             </h1>
                             <h1 className={styles.tagline}>
-                                <span>Apify SDK</span> for <span>JavaScript</span><br /> is a <span>toolkit</span> for<br /> building <span>actors</span>
+                                <span>Apify SDK</span> for <span>JavaScript</span> and <span>Node.js</span>
```
B4nan, line 23: the `<h1>` on like 19 needs to be changed in the same way (but without the `<span>` tags), this is duplicated to achieve that gradient (only for some words, yet with a unified gradient center)

B4nan (changes requested):

website/src/pages/index.js
jancurn: fixed

B4nan (approved):

PR #223, opened 2023-08-31
chore: set `onlyPurgeOnce` to `true` explicitly for `purgeDefaultStorages`

foxt451:
This is a step towards changing the default behaviour of `purgeDefaultStorages` to clear storages each time it gets called.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
         "@typescript-eslint/eslint-plugin": "^6.2.0",
         "@typescript-eslint/parser": "^6.2.0",
         "commitlint": "^17.6.7",
-        "crawlee": "^3.4.2",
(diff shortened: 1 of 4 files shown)


B4nan (approved):

package.json
```diff
@@ -65,7 +65,7 @@
         "@typescript-eslint/eslint-plugin": "^6.2.0",
         "@typescript-eslint/parser": "^6.2.0",
         "commitlint": "^17.6.7",
-        "crawlee": "^3.4.2",
+        "crawlee": "^3.5.3",
```
B4nan, line 68: FYI this was wrong, the place that needs the new range is the apify package (`packages/apify/package.json`), not the root deps (as that is basically just dev deps)

PR #227, opened 2023-09-08
feat: Use `.reboot()`  instead of `.metamorph()` for reboot

### LINKED ISSUE #220: Edit the `.reboot()` to use the new reboot endpoint
Edit the `.reboot()` method in the `Actor` class so that it uses the new Actor reboot endpoint. (will be done via a new method in the client)

jirimoravcik:
Closes #220

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -457,6 +457,10 @@ export class Actor<Data extends Dictionary = Dictionary> {
 
         const runId = this.config.get('actorRunId')!;
         await this.apifyClient.run(runId).reboot();
+
+        // Wait some time for container to be stopped.
+        const customAfterSleepMillis = this.config.get('metamorphAfterSleepMillis');
+        await sleep(customAfterSleepMillis);
     }
 
     /**
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -447,7 +447,7 @@ export class Actor<Data extends Dictionary = Dictionary> {
             return;
         }
 
-        // Waiting for all the listeners to finish, as `.metamorph()` kills the container.
+        // Waiting for all the listeners to finish, as `.reboot()` kills the container.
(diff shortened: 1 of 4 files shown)


B4nan (approved):

packages/apify/src/actor.ts
```diff
@@ -447,16 +447,16 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
             // `persistState` for individual RequestLists, RequestQueue... instances to be persisted
             ...this.config.getEventManager().listeners(EventType.PERSIST_STATE).map((x) => x()),
             // `migrating` to pause Apify crawlers
             ...this.config.getEventManager().listeners(EventType.MIGRATING).map((x) => x()),
         ]);
 
-        const actorId = this.config.get('actorId')!;
-        await this.metamorph(actorId);
+        const runId = this.config.get('actorRunId')!;
+        await this.apifyClient.run(runId).reboot();
```
fnesveda, line 459: We should sleep here after the API call, like we do with metamorph, so that the Actor waits for the reboot to happen (otherwise it could finish before the reboot happens).

fnesveda (changes requested):
One note about the reboot behavior

packages/apify/src/actor.ts
jirimoravcik: you're right, missed that, thanks

packages/apify/src/actor.ts
```diff
@@ -441,22 +441,26 @@ export class Actor<Data extends Dictionary = Dictionary> {
      *
      * @ignore
      */
-    async reboot(): Promise<void> {
+    async reboot(options: RebootOptions = {}): Promise<void> {
```
jancurn, line 444: Pls can we describe the use case of reboot() in jsdoc here? E.g. if the actor gets into irrecoverable error state, it can just reboot itself and continue work where it left off.

PR #228, opened 2023-09-08
docs: Improve Actor reboot docs with a use case

jirimoravcik:
Sry @jancurn I merged the previous PR before I noticed your commen


jancurn (approved):
Thanks

B4nan (approved):

PR #233, opened 2023-10-03
chore: support got-scraping v4

vladfrangu:
Needs https://github.com/apify/crawlee/pull/2110


vladfrangu:
@B4nan SIGSEGV on tests, looks like we need to move this repo to vitest too 😄 😢

B4nan (approved):

PR #236, opened 2023-10-05
fix: add more logging to `Actor.init` and `Actor.exit`

B4nan:
Also adds timeout handler around `Actor.setStatusMessage()`.

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -564,7 +564,7 @@ export class Actor<Data extends Dictionary = Dictionary> {
             const run = await addTimeoutToPromise(
                 () => this.apifyClient.run(runId).get(),
                 1000,
-                'Setting status message timed out after 1s',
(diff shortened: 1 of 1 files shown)


packages/apify/src/actor.ts
```diff
@@ -543,12 +548,25 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
 
         const runId = this.config.get('actorRunId')!;
+
         if (runId) {
-            const run = await this.apifyClient.run(runId).get();
+            // just to be sure, this should be fast
+            const run = await addTimeoutToPromise(
+                () => this.apifyClient.run(runId).get(),
+                1000,
+                'Setting status message timed out after 1s',
```
barjin, line 567:
```suggestion
                'Getting the current run timed out after 1s',
```

barjin (approved):

PR #260, opened 2023-11-28
docs: add more values to ApifyEnv

### LINKED ISSUE #170: Docs: ApifyEnv type missing most properties
### Issue description

Docs page: 
https://sdk.apify.com/api/apify/class/Actor#getEnv

Return value
https://api.apify.com/v2/key-value-stores/emJruom398sXAWxCP/records/ENV

We don't have to type everything but definitely the core things with IDs, anything that is useful for devs

### Code sample

_No response_

### Package version

latest

### Node.js version

node 16

### Operating system

_No response_

### Actor or run link

_No response_

### I have tested this on the `next` release

_No response_

### Other context

_No response_

foxt451:
Basically, moved over all what was left from APIFY_ENV_VARS and searched through docs for known descriptions
Closes #170

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -1552,6 +1552,9 @@ export interface MainOptions extends ExitOptions, InitOptions {}
  * This object is returned by the {@apilink Actor.getEnv} function.
  */
 export interface ApifyEnv {
+    // TODO: if this interface hasn't been changed for a while, you
+    // are invited to check if there are any new env vars (refer to APIFY_ENV_VARS and ACTOR_ENV_VARS)
+
     /**
      * ID of the actor (ACTOR_ID)
      */
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -1553,19 +1553,24 @@ export interface MainOptions extends ExitOptions, InitOptions {}
  */
 export interface ApifyEnv {
     /**
-     * ID of the actor (ACTOR_ID)
+     * ID of the actor (APIFY_ACTOR_ID)
      */
     actorId: string | null;
 
     /**
-     * ID of the actor run (ACTOR_RUN_ID)
+     * ID of the actor run (APIFY_ACTOR_RUN_ID)
      */
     actorRunId: string | null;
 
     /**
-     * ID of the actor task (ACTOR_TASK_ID)
+     * ID of the actor task (APIFY_ACTOR_TASK_ID)
      */
     actorTaskId: string | null;
+
+    /**
+     * ID of the Actor build used in the run. (APIFY_ACTOR_BUILD_ID)
+     */
+    actorBuildId: string | null;
 
     /**
      * ID of the user who started the actor - note that it might be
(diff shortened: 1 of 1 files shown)


packages/apify/src/actor.ts
```diff
@@ -1553,20 +1553,25 @@ export interface MainOptions extends ExitOptions, InitOptions {}
  */
 export interface ApifyEnv {
     /**
-     * ID of the actor (ACTOR_ID)
+     * ID of the actor (APIFY_ACTOR_ID)
```
B4nan, line 1556: please revert those, i believe we want to keep the plain `ACTOR_` env vars (check the git history for more details)

foxt451: Yeah, took a look at previous PRs but still not sure what is the thumb rule for delienating these as there is a lot of overlap. Are ACTOR_ specific to actor and APIFY_ to platform/user acc? If so, I should then replace all APIFY_ prefixes if there is an equivalent with ACTOR_, right?

packages/apify/src/actor.ts
```diff
@@ -1580,32 +1585,100 @@ export interface ApifyEnv {
...
-     * in megabytes (ACTOR_MEMORY_MBYTES)
+     * in megabytes (APIFY_MEMORY_MBYTES)
      */
     memoryMbytes: number | null;
+
+    /**
+     * If set to 1, the web browsers inside the Actor should run in headless
+     * mode because there is no windowing system available. (APIFY_HEADLESS)
+     */
+    headless: 1 | null;
```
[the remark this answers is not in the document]
foxt451: Took a look at Lukas' snapshot and it's string "1" there even. I guess that's because this funciton only cares about integer conversion and base on the following list, and it's missing there

```
export declare const INTEGER_ENV_VARS: readonly ["ACTOR_MAX_PAID_DATASET_ITEMS", "ACTOR_MEMORY_MBYTES", "ACTOR_WEB_SERVER_PORT", "ACTOR_MAX_PAID_DATASET_ITEMS", "APIFY_CONTAINER_PORT", "APIFY_DEDICATED_CPUS", "APIFY_MEMORY_MBYTES", "APIFY_METAMORPH_AFTER_SLEEP_MILLIS", "APIFY_PERSIST_STATE_INTERVAL_MILLIS", "APIFY_PROXY_PORT", "APIFY_SYSTEM_INFO_INTERVAL_MILLIS"];
```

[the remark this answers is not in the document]
foxt451: regarding it being here, the return type is the mirror of APIFY_ENV_VARS which lists it. And there is no actor equivalent

packages/apify/src/actor.ts
B4nan: i believe everything that has `ACTOR_` variant should use that, not sure if there are some that don't

jancurn: For context, originally the env vars were called `APIFY_XXX`, but with the introduction of the Actor philosophy, we wanted to de-couple Actors from Apify, and hence redefined the env vars as non-Apify `ACTOR_XXX` instead (see [here](https://github.com/apify/actor-specs#environment-variables)). The Apify SDK should use those whenever possible, the old `APIFY_XXX` env vars are kept on the platform just for backwards compatability.

packages/apify/src/actor.ts
```diff
@@ -1606,6 +1611,73 @@ export interface ApifyEnv {
      * in megabytes (ACTOR_MEMORY_MBYTES)
      */
     memoryMbytes: number | null;
+
+    /**
+     * If set to "1", the web browsers inside the Actor should run in headless
+     * mode because there is no windowing system available. (APIFY_HEADLESS)
+     */
+    headless: string | null;
```
metalwarrior665, line 1619: Why aren't these booleans? We already parse Dates above so this object already has parsed env, it is not 1:1 raw strings

metalwarrior665 (changes requested):

packages/apify/src/actor.ts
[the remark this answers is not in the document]
metalwarrior665: Sorry, I mistakenly created new review. I wanted to suggest using booleans but now I realized this is only updating the types. And changing the values would be breaking change even though they should have been parsed as booleans.

metalwarrior665 (approved):
I occasionally run [this actor](https://console.apify.com/actors/oXTuyqGR803PjzDha/source) to check for new env vars. I guess type getting out of date will be persistent problem as new env vars are being added on the platform and we automatically camel-case them. 

Can you add a TODO comment there to periodically update the type?

PR #272, opened 2024-01-08
feat: ignore proxy configuration locally if no valid token or password is found

### LINKED ISSUE #262: Better error handling and logging for Apify proxy configuration
### Feature

We want to change the behavior of the instance of `Actor.createProxyConfiguration`

Current behavior:
-  throws an error if a user is not logged in (doesn't provide the proxy password)
- throws an error if the user doesn't have access to the Apify proxy

Suggested change:
We would like to just print a warning to the console instead (or create some safer `createProxyConfiguration`) on the local environment. The Apify proxy just wouldn't be used at all, the configuration would do nothing.

Also, we proposed adding an additional warning - if the user has access to the Apify proxy but can't use it locally (this must be explained properly - free plan users).

We also propose an improved copy for the original error.
- It must be clear from the first error user can simply use CLI `apify login` cmd to get the proxy password.

### Motivation

Make Apify templates as seamless as possible. Now, a developer has to think about where it can be run and catch different errors. It is not clear from the error what to do and we have feedback from real users that struggled with this for quite some time.

Make `Actor.createProxyConfiguration` with warnings instead of error for local environment.

_No response_

### Other context

_No response_

B4nan:
We use the proxy configuration in all the templates, as we want users to run behind a proxy all the times. This PR changes the internal mechanism to gracefully ignore the proxy configuration locally when we don't see a valid token or proxy password. This way, new users can run their actors locally without being logged in.

This is how a run without proxy password look like:

```
INFO  System info {"apifyVersion":"3.1.14","apifyClientVersion":"2.8.4","crawleeVersion":"3.7.1","osType":"Darwin","nodeVersion":"v20.9.0"}
WARN  ProxyConfiguration: No proxy password or token detected, running without proxy. To use Apify Proxy locally, provide options.password or "APIFY_PROXY_PASSWORD" environment variable. If you add the "APIFY_TOKEN" environment variable, the password will be automatically inferred.
WARN  ProxyConfiguration: Invalid password provided: User not found.
INFO  CheerioCrawler: Starting the crawler.
INFO  CheerioCrawler: enqueueing new URLs
INFO  CheerioCrawler: Crawlee · Build reliable crawlers. Fast. | Crawlee {"url":"https://crawlee.dev/"}
INFO  CheerioCrawler: enqueueing new URLs
...
```

Closes #262

### DIFF
--- a/packages/apify/src/proxy_configuration.ts
+++ b/packages/apify/src/proxy_configuration.ts
@@ -331,6 +331,7 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
                     return;
                 }
             }
+
             const { password } = proxy!;
 
             if (this.password) {
@@ -380,12 +381,10 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
         // This is because the user might not have set up things correctly yet.
         // It still fails on the platform, where we don't want to allow this behavior.
         if (Actor.isAtHome()) {
-            this._throwApifyProxyConnectionError(connectionError);
-        } else {
-            this.log.warning(connectionError);
-            return false;
+            throw new Error(connectionError);
         }
 
+        this.log.warning(connectionError);
         return false;
     }
 
@@ -411,14 +410,6 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
         }
 
         return undefined;
-    }
-
-    /**
-     * Throws Apify Proxy is not connected
-     * @internal
-     */
-    protected _throwApifyProxyConnectionError(errorMessage: string) {
-        throw new Error(errorMessage);
     }
 
     /**
--- a/packages/apify/src/proxy_configuration.ts
+++ b/packages/apify/src/proxy_configuration.ts
@@ -7,6 +7,7 @@ import {
     ProxyConfiguration as CoreProxyConfiguration,
 } from '@crawlee/core';
 import { gotScraping } from '@crawlee/utils';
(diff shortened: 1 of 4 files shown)


packages/apify/src/proxy_configuration.ts
```diff
@@ -339,17 +360,33 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
...
+        }
+
+        // Throw only on the platform, locally we just print a warning and run requests without the proxy.
+        // This is because the user might not have set up things correctly yet.
+        // It still fails on the platform, where we don't want to allow this behavior.
+        if (Actor.isAtHome()) {
+            this._throwApifyProxyConnectionError(connectionError);
+        } else {
+            this.log.warning(connectionError);
+            return false;
```
barjin, line 386:
```suggestion
```

I still feel like there are way too many `return`s here... this one we can kill for sure though :)

jbartadev (approved):
Looks good 🤖

mnmkng:
CleanShot 2024-01-08 at 17 19 52@2x

I like this change! But do we need the double warning though? The second log is kinda pointless after the first one. Also a bit confusing. So is it a missing password or invalid password?

B4nan:
Yeah, I also don't like that. I can try to get rid of it, but it will make the code much uglier, that's why I haven't done it already :D 

I guess I can save a bool flag when the first warning is printed and ignore the other check.

PR #282, opened 2024-02-16
chore: bump `apify-client`

barjin:
as discussed in https://github.com/apify/crawlee/pull/2339

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3827,28 +3827,20 @@
             "link": true
         },
         "node_modules/apify-client": {
-            "version": "2.8.6",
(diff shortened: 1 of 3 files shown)


B4nan:
right, crawlee needs to go first as that is not dependent on the sdk or the client (only the e2e tests are)

so let's merge https://github.com/apify/crawlee/pull/2339 and do the release first. i'd do a minor/feature one.

B4nan:
And we should merge https://github.com/apify/crawlee/pull/2340 first before we create the docs snapshots, but I'll handle that myself, it's not blocking the release itself.

PR #293, opened 2024-04-09
docs(actor-scraper): use RequestQueueV2 for all Actor scrapers

drobnikj:


B4nan:
this is just a docs update, right? the actual changes #288, and this should have `docs:` type instead?

also the crawlee bump is to 3.8.2, and we should ship 3.9 today and update them right ahead ideally

drobnikj:
@B4nan True, I just forgot to add a changelog in https://github.com/apify/apify-sdk-js/pull/288

PR #301, opened 2024-05-23
feat: validate crawlee versions in `Actor.init`

### LINKED ISSUE #237: Warn about SDK vs crawlee version mismatch
Since the SDK depends on crawlee, users might configure their dependencies so those two don't fit each other, which ends up with multiple `@crawlee/core` installs, which in turn might cause weird issues, since there will be two global configs. This could potentially result in the crawlee storage methods (like `Dataset.pushData`) working with the memory storage even when on platform.

We should validate this in the `Actor.init` call, there needs to be a single crawlee installation.

B4nan:
The SDK depends on `@crawlee/core`, but if the user installs an incompatible version of crawlee, they might end up with two installations of the `@crawlee/core` which means multiple static registers like the one for the global configuration, which breaks the integration, since the SDK will only touch the global config it sees, but it's a different one than the explicitly installed package.

Closes #237

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17095,6 +17095,7 @@
                 "@crawlee/types": "^3.9.0",
                 "@crawlee/utils": "^3.9.0",
                 "apify-client": "^2.9.0",
+                "fs-extra": "^11.2.0",
                 "ow": "^0.28.2",
                 "semver": "^7.5.4",
                 "tslib": "^2.6.2",
--- a/packages/apify/package.json
+++ b/packages/apify/package.json
@@ -63,6 +63,7 @@
         "@crawlee/types": "^3.9.0",
         "@crawlee/utils": "^3.9.0",
         "apify-client": "^2.9.0",
+        "fs-extra": "^11.2.0",
         "ow": "^0.28.2",
         "semver": "^7.5.4",
         "tslib": "^2.6.2",
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -41,7 +41,7 @@ import { KeyValueStore } from './key_value_store';
 import { PlatformEventManager } from './platform_event_manager';
 import type { ProxyConfigurationOptions } from './proxy_configuration';
 import { ProxyConfiguration } from './proxy_configuration';
-import { logSystemInfo, printOutdatedSdkWarning } from './utils';
+import { checkCrawleeVersion, logSystemInfo, printOutdatedSdkWarning } from './utils';
 
 /**
  * `Actor` class serves as an alternative approach to the static helpers exported from the package. It allows to pass configuration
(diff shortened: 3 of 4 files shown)


vladfrangu:
Why do we not just declare crawlee as a peer dependency? Would solve this, no?

B4nan:
You mean declare it only as a peer instead of a real dependency? In the good old times, that would require the user to install it. Nowadays I believe different package managers do different things, as NPM just decided to install them (which is IMO nonsense in the first place), but yarn wont do that and fail. We don't want people to install both if they don't want to use crawlee.

And honestly, I would like to keep this validation even if it would help.

vladfrangu:
I'm just worried about relying on fs checks for this... What if

- the users use yarn/pnpm pnp
- the users use pnpm (will it support the hardlinks? Probably)
- <some package manager> doesn't respect that folder structure and flattens it out (like most of them do)

Specifically I'm worried that if our checks don't cover enough we'll need to patch release a few times till we get it right™️ 

Peer dependencies _would_ solve _some_ of these issues, but because the ecosystem cannot *decide what the hell it wants*, it's like you said, a mess. npm now automatically installs peer deps (but if you specify a different version that matches the range it won't install twice), yarn does... either nothing or something (looking at you [peerDependencyMetadata](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta)), who knows about pnpm/bun.

There is also [`optionalDependencies`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies) but not sure if that'll do what we want either. 😢 

---

Unrelated, but please use path.join for paths, something something windows 😄

B4nan:
> doesn't respect that folder structure and flattens it out (like most of them do)

Flattening works only if the version ranges match, and then there is no problem.

I've been using similar checks for quite some time (years), and I got inspired by docusaurus doing the same (or similar) thing too. Yes, we might need to improve the logic over time, but only to make it more strict, not to fix some issues it introduces.

> There is also [optionalDependencies](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies) but not sure if that'll do what we want either. 😢

This is not an optional peer, its required, that's why its an actual dependency in the first place.

vladfrangu:
> This is not an optional peer, its required, that's why its an actual dependency in the first place

Optionals still get installed. They're "optional" in the sense that if it fails to install it won't throw an error and abort the whole install process

B4nan:
> Unrelated, but please use path.join for paths, something something windows 😄

used `path.normalize`, that should do the job too, right?

B4nan:
> Optionals still get installed. They're "optional" in the sense that if it fails to install it won't throw an error and abort the whole install process

Well, that's the part I was referring to, NPM does that nowadays, I don't think yarn would (at least it wasn't doing that some time ago and failed instead)...

vladfrangu:
> > Unrelated, but please use path.join for paths, something something windows 😄
> 
> used `path.normalize`, that should do the job too, right?

I think so, yeah!

B4nan:
2bc86c71d1cd378cd48410383a4849e578256a30 will help with pnpm and yarn pnp

PR #306, opened 2024-06-14
Update RequestListOptions.md

jancurn:


B4nan (approved):
also fixed it in crawlee docs, as that is where the `RequestListOptions` are now documented

https://github.com/apify/crawlee/commit/5e1b5eafe0705328d302e269892e5e3ca42ef278

PR #309, opened 2024-07-03
chore: bump client version

barjin:
Following the latest [Apify client release](https://github.com/apify/apify-client-js/releases/tag/v2.9.4), we want to ensure we get the new [API error fields](https://github.com/apify/apify-client-js/commit/3b7c4b73607f40a74f31f39491025c63bc92e40e) via the SDK.
I don't know the details, this is mostly on @gippy 's request. 

It also made me realize that I made a small semver mistake with the latest Apify client release (there are new "features"(?), it should have probably been a `minor` version). Do we want to address this somehow?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3987,9 +3987,9 @@
             "link": true
         },
         "node_modules/apify-client": {
-            "version": "2.9.0",
-            "resolved": "https://registry.npmjs.org/apify-client/-/apify-client-2.9.0.tgz",
-            "integrity": "sha512-QyTyDjGXQV5Aey3Jbs5pxDIaJmzTSkByYGf6JambHkKHuO/qR+3zXH6ENABzcleeGAkc73PCg24F3nYFXqhNHw==",
+            "version": "2.9.4",
+            "resolved": "https://registry.npmjs.org/apify-client/-/apify-client-2.9.4.tgz",
+            "integrity": "sha512-Cot4fXspaTVne2CHIQoMNePGUxeMPGQkSRSWOSnQwycbfyFJJ73gdgTlyh5PhmwVz9EXmwp9xw2wgSKeULZHUg==",
             "dependencies": {
                 "@apify/consts": "^2.25.0",
                 "@apify/log": "^2.2.6",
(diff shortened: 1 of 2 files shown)


gippy:
> Following the latest [Apify client release](https://github.com/apify/apify-client-js/releases/tag/v2.9.4), we want to ensure we get the new [API error fields](https://github.com/apify/apify-client-js/commit/3b7c4b73607f40a74f31f39491025c63bc92e40e) via the SDK. I don't know the details, this is mostly on @gippy 's request.
> 
> It also made me realize that I made a small semver mistake with the latest Apify client release (there are new "features"(?), it should have probably been a `minor` version). Do we want to address this somehow?

We discussed it today with Martin and in the end it might not be needed, sorry. I will try out the current version if everything works and if yes than this might later be replaced with other change.

B4nan:
Yeah, I would only raise the version range once we actually use something from the new APIs, for testing it should be enough already.

> Do we want to address this somehow?

Let's keep it as is, we don't need to ship feature releases for small "features" :] Maybe we should automate this (and remove versioned docs for the client, as it's not adding much value anyway), but let's leave that for later.

barjin:
Alright then, closing this PR 👍🏽

gippy:
@B4nan So I tested it and I'm not sure if it behaves as you said.

I have this package.json and based on my discussion the latest version of SDK should automatically install client version 2.9.4:
CleanShot 2024-07-08 at 14 07 23@2x

But when the run starts after build it starts with this log message:
CleanShot 2024-07-08 at 14 08 08@2x

Where the client version is still 2.9.3 and the error from validation does not contain the data field. It's in an actor without package-lock, so it should not be blocked this way.

B4nan:
Try to have a lock file there, then it should be easy to test it locally. I can imagine without the lockfile it will just use the version preinstalled in the base docker image. You can also use `overrides` to set the dependency of the SDK from outside explicitly.

gippy:
I cannot easily try it localy without complicated hacks. It's not released in API yet, but we can do the release and then test it locally. 

But we need this to work for all users automatically without any hacks.
Ideally we would just update the templates to use new version of SDK and SDK will use the version of client with this change, where you can access the data of the error.

B4nan:
I am rather sure it's about the lockfile (and without that you are probably getting the version installed in the base docker images). If you create new template locally, it will have the latest client installed.

image

PR #310, opened 2024-07-04
feat: add standby port to configuration

jirimoravcik:
This PR adds Actor Standby port to the configuration.


fnesveda (approved):

packages/apify/src/configuration.ts
```diff
@@ -148,6 +149,7 @@ export class Configuration extends CoreConfiguration {
         ACTOR_INPUT_KEY: 'inputKey',
         ACTOR_MEMORY_MBYTES: 'memoryMbytes',
         ACTOR_RUN_ID: 'actorRunId',
+        [ACTOR_ENV_VARS.STANDBY_PORT]: 'standbyPort',
```
B4nan, line 152: I would rather do this to be consistent

```suggestion
        ACTOR_STANDBY_PORT: 'standbyPort',
```

B4nan (approved):
just one nit but i don't mind merging as is

Issue #322, opened 2024-09-02
SDK does not output proper ESM


ryanleecode:
### Feature

Only the index.mjs file is ESM, and it just re-exports everything which is CJS. this is wrong.

### Motivation

.

### Ideal solution or implementation, and any additional constraints

https://github.com/isaacs/tshy

PR #323, opened 2024-09-03
fix: esm builds

ryanleecode:
#322


B4nan:
Thanks for the PR, but we won't be merging this one, definitely not in such shape. If we are to change the monorepo setup, we will handle it ourselves, we are not interested in a complete rewrite like this. Also I hope you don't expect us to merge a PR that is doing so many things at once, including changing package manager, and doing many breaking changes like changing the TS configs completely.

I will leave it open, as I feel like parts of it might be usable at least as an inspiration.

Note that we are considering going ESM native in next major instead of dealing with dual builds (especially since there is now the experimental support for require(esm) in both node 20 and 22). It will likely happen later this year.

PR #326, opened 2024-10-04
fix: decode special characters in proxy `username` and `password`

B4nan:
Related: https://github.com/apify/crawlee/issues/2696


barjin (commented):
If we're to assume that `this.password` is un-encoded, then we probably should encode it here:

https://github.com/apify/apify-sdk-js/blob/449767178adc4cf4128b955e70be0988fbf7a765/packages/apify/src/proxy_configuration.ts#L355-L360

E.g. the `URL` class encodes `.password` (and `.username` etc) assignments automatically.

barjin:
Other than that, lgtm based on the other PR 👍🏽

PR #328, opened 2024-10-08
feat: add standby URL, change default port

jirimoravcik:
This PR introduces a new environment variable and also changes the default number of standby port to 4321.

### DIFF
--- a/packages/apify/src/configuration.ts
+++ b/packages/apify/src/configuration.ts
@@ -21,6 +21,9 @@ export interface ConfigurationOptions extends CoreConfigurationOptions {
     proxyPassword?: string;
     proxyPort?: number;
     proxyStatusUrl?: string;
+    /**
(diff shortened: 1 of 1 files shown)


B4nan (approved):

packages/apify/src/configuration.ts
```diff
@@ -22,6 +22,7 @@ export interface ConfigurationOptions extends CoreConfigurationOptions {
     proxyPort?: number;
     proxyStatusUrl?: string;
     standbyPort?: number;
```
fnesveda, line 24: Is it possible to mark this property as deprecated?

fnesveda (approved):
Looks good, just one quick question

packages/apify/src/configuration.ts
jirimoravcik: done, added.

PR #334, opened 2024-10-30
fix: remove zwsp characters from markdown docs

barjin:
Removes unnecessary `zero-width space` characters from the `mdx` files. The source of those is unclear - git blame says that they have been there since > 2 years ago. The on-page rendering remains the same with or without those characters.

Closes https://github.com/apify/apify-docs/issues/1262


B4nan:
I recall this was needed when the component was the first thing on the line, but maybe things changed in v3.

barjin:
You were once again faster 👍🏽  thanks for the clarification

PR #335, opened 2024-11-01
feat: add SDK and Crawlee version to the `ApifyClient`'s user agent

### LINKED ISSUE #331: Add SDK version to Client user agent
### Feature

It would be great if we could add information about SDK to Apify Client calls that are being called with Apify SDK and not Client directly.

### Motivation

We would like to understand which Actor runs on the platform are triggered from Client (external calls) or SDK (implies either some Actor implementation, or perhaps a wrong selection of library). The user agents are already tracked on platform, so no changes needed there.

The best way to do that would probably be by adding SDK version to the Client user agent.

_No response_

### Other context

If this is accepted, it might be worth adding even more info, like Crawlee version, if called from Crawlee.

B4nan:
The support for this was added in client v2.10 (https://github.com/apify/apify-client-js/pull/602/files#diff-c33bae26ce4a7b3559aef1321052b4e22c099a786ae6d660d9409bee032c7fa8R44), user agent looks like this:

```
ApifyClient/2.9.8 (Darwin; Node/v22.11.0); isAtHome/false; SDK/3.1.1; Crawlee/3.11.5
```

Closes #331

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3986,9 +3986,9 @@
             "link": true
         },
         "node_modules/apify-client": {
-            "version": "2.9.7",
-            "resolved": "https://registry.npmjs.org/apify-client/-/apify-client-2.9.7.tgz",
-            "integrity": "sha512-bI7sRCIlyt9M1qv2N0tCIui/jVQqf5FG6pZIfAUZrfWmHuE15QHp92kSjBK5aL+Iv5FDZ4zoondrCZNQSYFB/w==",
+            "version": "2.10.0",
+            "resolved": "https://registry.npmjs.org/apify-client/-/apify-client-2.10.0.tgz",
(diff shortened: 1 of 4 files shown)


B4nan:
@mnmkng This adds both SDK and Crawlee versions to the UA, since SDK uses code from crawlee, it's technically always there (and we validate that the version it uses is the same), do you think we should differentiate calls just from SDK and those from crawlee? I wouldn't care about that personally, especially since users can use the static `Actor...` methods in crawlee projects too, so what would be trying to differentiate in the first place? The only place crawlee uses the client internally is to communicate with the storages, that is handled via the `Actor.init` call, and that comes from the user (and will be there even without crawlee being in use explicitly), so I am not sure if/how could this be done.

B4nan:
I'll merge this in, we can always improve later.

PR #345, opened 2024-12-18
fix: prevent reboot loops

fnesveda:
We can't use `Actor.reboot()` inside a `migrating` event handler, because it internally calls all the handlers for the `migrating` event, so calling that on the `migrating` event would get into an infinite loop. This fixes it by allowing the reboot to be called only once.


janbuchar (approved):
LGTM, @B4nan should probably see this as well

B4nan (approved):

PR #346, opened 2024-12-23
feat: Actor.charge()

janbuchar:
- Depends on https://github.com/apify/apify-client-js/pull/623
- Draws inspiration from https://github.com/metalwarrior665/actor-charge-manager-poc

## TODO
- [x] write e2e tests


packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,230 @@
...
+            await this.chargingLogDataset.pushData({
+                eventName,
+                eventTitle: pricingInfo.title,
+                eventPriceUsd: pricingInfo.price,
+                timestamp,
+            });
+        }
+
+        return {
+            eventChargeLimitReached: this.calculateTotalChargedAmount() >= this.maxTotalChargeUsd,
```
metalwarrior665, line 156: 1. I think it is better to return `true` even if `this.calculateTotalChargedAmount() < this.maxTotalChargeUsd` and there is not enough budget for any event. That tells the developer that he can finish the scrape. Also, `this.calculateTotalChargedAmount() >= this.maxTotalChargeUsd` probably can never happen since you return early if we cannot charge at least 1 event so we never overflow the limit.
2. I was returning if there is budget for this specific event because that is what limits the `chargedCount`. Maybe we can return that too? Or an object with how many of each event can still be charged?

metalwarrior665 (commented):
Thanks! I don't have any major complain, just a few notes.

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,230 @@
...
+
+        // Keeping float precision issues at bay
+        return Number(result.toFixed(6));
+    }
+
+    /**
+     * How many events of a given type can still be charged for before reaching the limit;
+     * If the event is not registered, returns Infinity (free of charge)
+     */
+    private calculateEventChargeCountTillLimit(eventName: string): number {
```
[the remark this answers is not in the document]
janbuchar: 1. `ChargingManager` is internal, so the devs won't be able to access it without serious shenanigans anyway
2. it's not part of the spec (whitepaper)
3. it's always better to start small when designing interfaces

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        const timestamp = new Date().toISOString();
+
+        for (let i = 0; i < chargedCount; i++) {
+            await this.chargingLogDataset.pushData({
+                eventName,
+                eventTitle: pricingInfo.title,
+                eventPriceUsd: pricingInfo.price,
+                timestamp,
```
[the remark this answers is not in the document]
janbuchar: It's not part of the whitepaper

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * Charge for a specified number of events - sub-operations of the Actor.
+     *
+     * @param options The name of the event to charge for and the number of events to be charged.
+     */
+    async charge({ eventName, count = 1 }: ChargeOptions): Promise<ChargeResult> {
+        if (!this.isPayPerEvent) {
+            log.info('Ignored attempt to charge for an event - the Actor does not use the pay-per-event pricing');
```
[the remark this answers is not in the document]
janbuchar: No opinion, I can make it `debug`.

packages/apify/src/actor.ts
```diff
@@ -896,6 +921,29 @@ export class Actor<Data extends Dictionary = Dictionary> {
         return undefined;
     }
 
+    /**
+     * Charge for a specified number of events - sub-operations of the Actor.
+     *
+     * @param options The name of the event to charge for and the number of events to be charged.
+     */
+    async charge(options: ChargeOptions): Promise<ChargeResult> {
+        return this.chargingManager.charge(options);
```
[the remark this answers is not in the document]
janbuchar: Good point, can do.

packages/apify/src/internals/charging.ts
janbuchar: 1. it may be non-obvious to some what `{charged: 41.3, eventChargeLimitReached: true}` means if the limit is 42 and the lowest price we have is 1, what do you think about that? Also "probably" and "never" are two different things, it doesn't hurt to write conditions defensively.
2. it is possible, but the use case of an Actor that decides what kind of work it's going to do based on the budget sounds contrived at best and sketchy at worst. Is this backed up by some research/survey, or is it a just-in-case thing?

metalwarrior665: 1.
  a) `chargedCount` is the number of charged events in this charge, right? So there cannot be float price like in your example `{charged: 41.3, eventChargeLimitReached: true}`. I agree though that it is non-obvious, maybe we just need better naming. But I think it makes sense to tell the devs when they can no longer charge anything, that's what they care about and when they should finish, not exceeding the max (which they should not exceed anyway because it is a limit). Also, it would be consistent with the `if (chargedCount === 0) {` where you return `eventChargeLimitReached: true` even if the limit wasn't necessarily exceeded.
  b) `probably` is my bad, I should have thought through it properly. I was wrong, this condition can happen if `0 < chargedCount < count`. So I'm just back to the 1.a) point.

2. That already happens in some Actors. I agree it is contrived but the use-case to fill the budget "just enough" is real. Usually, the Actor will decide if they want to spend the budget breath-first (first get products and then add reviews) or depth-first (first product with all features) or they could even leave some budget if it is needed for final upload. I recommend doing depth-first always since it is most intuitive and straightforward. But eventually you run into a situation where you can charge only some events so you need to check that and decide what to enqueue. An alternative "best practice" approach would be to just charge until you exceed the limit and then finish. Pros of that is that it would be simpler to implement (both for devs and in SDK) and it would fill up the budget fully. Cons is that for the exceeding amount, devs would not get paid. When I think about it like that, it sounds pretty good.

Let's hear from others.

packages/apify/src/actor.ts
```diff
@@ -609,13 +617,30 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+        const chargeResult = eventName !== undefined
+            ? await this.charge({ eventName, count: Array.isArray(item) ? item.length : 1 })
+            : undefined;
+
         const dataset = await this.openDataset();
-        return dataset.pushData(item);
+
+        if (chargeResult?.eventChargeLimitReached) {
+            const items = Array.isArray(item) ? item : [item];
+            await dataset.pushData(items.slice(0, chargeResult.chargedCount));
```
metalwarrior665, line 636: @halvko had a good point that in the super rare case when the Actor crashes in between the `.charge` and `.pushData` calls, it would be better to have pushData first and only after finish the charge. Considering we are talking few milliseconds, this will be super rare and it will complicate the code a bit but it would technically be more correct.

test/e2e/runSdkTests.mjs
```diff
@@ -109,24 +117,54 @@ async function runTest(dirName) {
...
+                        {
+                            pricingModel: 'FREE',
+                        },
+                    ],
+                },
+            );
+        }
+
+        exitCode = testProcess.exitCode;
+    } else {
```
B4nan, line 159: if you first check the inverse and throw, you can remove the nesting

```ts
if (buildResult.status !== 'SUCCEEDED') {
    throw new Error('Build failed');
}

// no nesting needed...
```

and the `exitCode` can be a const that way

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
vladfrangu: Why not emit warning once? 👀

test/e2e/sdk/actorCharge/test.mjs
janbuchar, line 1: I'd be super grateful if anyone could help me brainstorm more cases to test.

test/e2e/runSdkTests.mjs
janbuchar: Not the biggest problem I had, but thanks for pointing it out! :grin:

packages/apify/src/internals/charging.ts
janbuchar: I believe I made `eventChargeLimitReached` behave more consistenly and also added the `chargeableWithinLimit` attribute to `ChargeResult`. What do you think?

packages/apify/src/actor.ts
janbuchar: My intuition would be that it's a bigger problem if we have items in dataset that were not in fact successfully charged than the opposite. Does anyone else want to chime in?

packages/apify/src/internals/charging.ts
metalwarrior665: Yeah, this works well for us

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
metalwarrior665: cc @mhamas This is not blocking this PR but something to think about.

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
metalwarrior665: The `chargeableWithinLimit` should mostly cover this use-case

packages/apify/src/actor.ts
metalwarrior665: My intuition is that customer is much more likely to be pissed off that they paid for no value, than a developer that would provide free data for a customer :)

test/e2e/sdk/actorCharge/src/main.mjs
```diff
@@ -0,0 +1,12 @@
+import { Actor, log } from 'apify';
+
+await Actor.init();
+
+const chargeResult = await Actor.charge({
+    eventName: 'foobar',
+    count: 4,
+});
```
metalwarrior665, line 8: We could try to utilize the chargeResult in some way. The best that comes to my mind is to store it to KV store and then just load that and check the values

packages/apify/src/actor.ts
janbuchar: Ah, I lost context. I thought I was looking at the code responsible for the charge-audit dataset :facepalm:. You are correct, I'll try and fix this.

package.json
```diff
@@ -40,7 +40,10 @@
         "build": "turbo run build",
         "ci:build": "turbo run build --cache-dir=\".turbo\"",
         "test": "vitest run --silent",
-        "test:e2e": "node test/e2e/run.mjs",
+        "test:e2e": "npm run test:e2e:scrapers && npm run test:e2e:sdk",
+        "test:e2e:scrapers": "node test/e2e/runScraperTests.mjs",
+        "test:e2e:sdk": "npm run test:e2e:sdk:tarball && node test/e2e/runSdkTests.mjs",
+        "test:e2e:sdk:tarball": "cd packages/apify; mv $(npm pack | tail -n 1) ../../test/e2e/apify.tgz",
```
B4nan, line 46: do you really want `;` here and not `&&`?

also, how exactly does this work?

packages/apify/src/actor.ts
```diff
@@ -222,6 +227,9 @@ export class Actor<Data extends Dictionary = Dictionary> {
         log.debug(`Default storages purged`);
 
         Configuration.storage.enterWith(this.config);
+
+        await this.chargingManager.init();
+        log.debug(`ChargingManager initialized`);
```
B4nan, line 232: nit: shouldn't this be inside the `init` method?

packages/apify/src/actor.ts
```diff
@@ -1512,6 +1580,36 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+    static getChargedEventCount(eventName: string): number {
+        return Actor.getDefaultInstance().getChargedEventCount(eventName);
+    }
+
+    /**
+     * Get information about the pricing for this Actor.
+     */
+    static getPricingInfo(): ActorPricingInfo {
+        return Actor.getDefaultInstance().getPricingInfo();
+    }
```
B4nan, line 1611: Do we really want those methods on the `Actor` class directly? Maybe better to expose the charging manager instead?

I remember the pushback when I added `getInputOrThrow` some time ago, I think this was supposed to be following the Actor specs? cc @netmilk @jancurn

packages/apify/src/actor.ts
```diff
@@ -1304,9 +1366,15 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
      */
-    static async pushData<Data extends Dictionary = Dictionary>(item: Data | Data[]): Promise<void> {
-        return Actor.getDefaultInstance().pushData(item);
+    static async pushData<Data extends Dictionary = Dictionary>(item: Data | Data[]): Promise<void>;
+    static async pushData<Data extends Dictionary = Dictionary>(item: Data | Data[], eventName: string): Promise<ChargeResult>;
+    static async pushData<Data extends Dictionary = Dictionary>(item: Data | Data[], eventName?: string): Promise<ChargeResult | void> {
+        if (eventName === undefined) {
+            return await Actor.getDefaultInstance().pushData(item);
+        }
+        return await Actor.getDefaultInstance().pushData(item, eventName);
```
B4nan, line 1377: why the `await`s? and why not just `pushData(item, eventName)` for both cases? its effectively the same, no? missing param will give you `undefined`, this will only differ when you use `arguments` property

---

right, because of the overloads. cant help myself, but i don't like overloads at all. if you want to have them here, you need to document all of them separately, otherwise only the first one will be properly documented in our API docs

i would consider a new method for this instead of overloads, but then again, I am not sure if we want to pollute this `Actor` class more, it always has tons of methods

packages/apify/src/actor.ts
```diff
@@ -609,13 +617,29 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
         this._ensureActorInit('pushData');
 
         const dataset = await this.openDataset();
-        return dataset.pushData(item);
+
+        const maxChargedCount = eventName !== undefined ? this.chargingManager.calculateMaxEventChargeCountWithinLimit(eventName) : Infinity;
+        const toCharge = Array.isArray(item) ? item.length : 1;
+
+        if (toCharge > maxChargedCount) {
+            const items = Array.isArray(item) ? item : [item];
```
B4nan, line 634: I would maybe add `asArray` helper to simplify this, now you do it twice in the same method.

packages/apify/src/actor.ts
```diff
@@ -609,13 +617,29 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+            await dataset.pushData(items.slice(0, toCharge));
```
B4nan, line 635: this is just an array copy, right? but far less readable

```suggestion
            await dataset.pushData([...items]);
```

or should this push only up to the limit? now it seems to use `item.length` so pushing everything

packages/apify/src/actor.ts
```diff
@@ -609,13 +617,29 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+        } else {
+            await dataset.pushData(item);
+        }
+
+        if (eventName) {
+            return await this.chargingManager.charge({ eventName, count: Math.min(toCharge, maxChargedCount) });
```
B4nan, line 641:
```suggestion
            return this.chargingManager.charge({ eventName, count: Math.min(toCharge, maxChargedCount) });
```

packages/apify/src/configuration.ts
```diff
@@ -185,6 +189,7 @@ export class Configuration extends CoreConfiguration {
         standbyPort: +LOCAL_ACTOR_ENV_VARS[ACTOR_ENV_VARS.STANDBY_PORT],
         metamorphAfterSleepMillis: 300e3,
         persistStateIntervalMillis: 60e3, // This value is mentioned in jsdoc in `events.js`, if you update it here, update it there too.
+        testPayPerEvent: false,
```
B4nan, line 192: do we have this documented somewhere? how does the testing work?

package.json
janbuchar: It's very unlikely that the `cd` would fail, so... `&&` is correct, but it's bordeline bikeshedding.

What happens is that `test:e2e:sdk:tarball` builds an npm-install-able .tgz file in the `test/e2e` directory. Then it is copied into each Actor build and picked up and installed in the Dockerfile.

packages/apify/src/actor.ts
janbuchar: No idea, I have no problem adjusting this. AFAIK only `Actor.charge` is described in the whitepaper (https://whitepaper.actor/#charging-money)

package.json
B4nan: Well, I had to look up what `;` does, unlike `&&` which we all use pretty much everywhere. That's what triggered this comment mainly, and when I read what `;` does, it just feels wrong to use it. Agreed its very unlikely to cause anything else then confusion, but that's also a problem.

> What happens is that test:e2e:sdk:tarball builds an npm-install-able .tgz file in the test/e2e directory. Then it is copied into each Actor build and picked up and installed in the Dockerfile.

We should probably adopt this in the crawlee e2e tests too, that weird script for copying packages haunts me every time I deal with the e2e tests there.

packages/apify/src/actor.ts
janbuchar: That was a bug - it is only supposed to push `maxChargedAmount`. Thanks

packages/apify/src/actor.ts
janbuchar: I mean, sure. Does this make any practical difference, other than making the line shorter?

packages/apify/src/configuration.ts
janbuchar: Same place as the rest of PPE support in the SDK - in the future.

You're supposed to set the `ACTOR_TEST_PAY_PER_EVENT` environment variable when developing locally to put the ChargingManager into PPE pricing mode. If you don't do that, it will only log a warning when you call `Actor.charge` or something.

package.json
janbuchar, line 46:
```suggestion
        "test:e2e:sdk:tarball": "cd packages/apify && mv $(npm pack | tail -n 1) ../../test/e2e/apify.tgz",
```

packages/apify/src/actor.ts
netmilk: I feel the Actor namespace and the domain term are overloaded already, I'm in favor of organizing the non-essential (non-whitepaper, convenience) functionality in a nested/separate namespace.

@B4nan Would you mind sharing a snippet of how you would access the "charging manager", please?

B4nan: We would need a getter for it, now it's private, e.g. `Actor.getChargingManager().foo()`.

packages/apify/src/actor.ts
janbuchar: Sure, I can document those overloads separately. In this case, adding a parameter to `pushData` seemed like a less invasive solution to me, but a `pushDataAndCharge` method is also an option :shrug:

packages/apify/src/actor.ts
janbuchar: Cool, thanks for the explanation!

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,281 @@
...
+                eventChargeLimitReached: true,
+                chargedCount: 0,
+                chargeableWithinLimit: calculateChargeableWithinLimit(),
+            };
+        }
+
+        const pricingInfo = this.pricingInfo[eventName] ?? {
+            price: 0,
+            title: 'Unknown event',
+        };
```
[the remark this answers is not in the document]
janbuchar: Yes, for local development where we don't have any information about the pricing. In that case, the price should probably be zero so that it is possible to reach the maximum budget.

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,281 @@
...
+                    chargeCount,
+                    totalChargedAmount: chargeCount * (this.pricingInfo[eventName]?.price ?? 0),
+                };
+            }
+        }
+
+        if (this.isPayPerEvent) {
+            // Set up charging log dataset
+            if (this.isAtHome) {
+                const datasetId = await (async () => {
```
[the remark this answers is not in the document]
B4nan: You can just wrap the code in `{ ... }` to achieve the same, no need for IIFE. 

I wouldn't care about this personally, but I don't mind it. Better than async IIFE, that just feels weird outside of some `main.js` script in CJS projects.

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,281 @@
...
+    private maxTotalChargeUsd: number;
+    private isAtHome: boolean;
+    private actorRunId: string | undefined;
+    private pricingModel: string | undefined = undefined;
+    private purgeChargingLogDataset: boolean;
+    private notPpeWarningPrinted = false;
+
+    private pricingInfo: Record<string, {price: number; title: string}> = {};
+    private chargingState: Record<string, ChargingStateItem> | undefined = undefined;
+    private chargingLogDataset: Dataset | undefined;
```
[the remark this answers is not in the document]
janbuchar: Yeah, I'll bring in biome in a separate PR later.

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
janbuchar: But the result of the code block cannot be assigned to `datasetId`, right? That's the problem that the do-expression proposal is trying to solve. I'll just make a private method I guess.

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,281 @@
...
+        this.isAtHome = configuration.get('isAtHome');
+        this.actorRunId = configuration.get('actorRunId');
+        this.purgeChargingLogDataset = configuration.get('purgeOnStart');
+
+        if (configuration.get('testPayPerEvent') === true) {
+            if (this.isAtHome) {
+                throw new Error('Using the ACTOR_TEST_PAY_PER_EVENT environment variable is only supported in a local development environment');
+            }
+
+            this.pricingModel = 'PAY_PER_EVENT';
```
mhamas, line 40: Suggestion: it'd be safer not to use `string` type for for `pricingModel`, but instead create a dedicated type. Passing pure strings around us error-prone.

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /**
+     * Initialize the ChargingManager by loading pricing information and charging state via Apify API.
+     */
+    async init(): Promise<void> {
+        this.chargingState = {};
+
+        // Retrieve pricing information
+        if (this.isAtHome) {
```
mhamas, line 57: Q: what does exactly `isAtHome` mean? Would be good to document, it's non-obvious to me.

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,281 @@
...
+            const run = (await this.apifyClient.run(this.actorRunId!).get())!;
```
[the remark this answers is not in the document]
mhamas: Q: isn't the `apifyClient` typesafe? Why the TS doesn't shout when you are passing potentially `undefined` to `apifyClient.run`?

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,281 @@
...
+            if (this.isAtHome) {
+                const datasetId = await (async () => {
+                    const defaultStore = await KeyValueStore.open();
+
+                    const storedDatasetId = await defaultStore.getValue<string>(this.PLATFORM_CHARGING_LOG_DATASET_ID_KEY);
+                    if (storedDatasetId !== null) {
+                        return storedDatasetId;
+                    }
+
+                    const dataset = await this.apifyClient.datasets().getOrCreate();
```
mhamas, line 93: Q: Are you always creating this dataset? Seems you are. If that's the case, maybe this should be part of the platform itself? Who owns this dataset and is charged for its storage? How will it work with scoped tokens in the future?

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
mhamas: +1 for private function

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    private calculateTotalChargedAmount(): number {
+        if (this.chargingState === undefined) {
+            throw new Error('ChargingManager is not initialized');
+        }
+
+        const result = Object.values(this.chargingState).map(({ totalChargedAmount }) => totalChargedAmount).reduce((sum, inc) => sum + inc, 0);
+
+        // Keeping float precision issues at bay
+        return Number(result.toFixed(6));
```
mhamas, line 235: Q: is this actually helpful? Wouldn't any float precision issues actually happen after the 6th decimal digit? I guess yes, in which case you are just rounding here and introducing potentially bigger discrepancies than the possible native floating point issues.

For context, I think 6 decimals should be more than enough, but I don't t think this is actually necessary here.

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
mhamas: As written in another comment, I have concerns about creating a special dataset just for logs (because of size, charging, retention, permissions). We actually discussed this with @mtrunkat and @fnesveda and we decided against that, and instead we store the system log elsewhere (although I was actually in pro-dataset camp :D, I think it was mostly @fnesveda against and Mara decided at the end). But if you've learnt by the practice that a log dataset is always necessary, maybe it should be an integral part of the platform itself? Let's discuss offline @metalwarrior665 ?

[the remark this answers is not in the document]
metalwarrior665: @janbuchar @B4nan @netmilk  Ok, we had a chat with @mhamas and there are several serious concerns with the debug dataset that didn't cross my mind previously.

1. Consider there can be use-cases where developer would charge per LLM token so each charge would be in thousands like `Actor.charge(llm-token, 4562)`. This would create a huge pressure on the dataset, namely 
    a) Pushing thousands takes some time
    b) push items cost for developer
    c) timed storage cost for user
2. I didn't fully realizes that after stripping the developer provided metadata for each event, the `chargingLogDataset` now only contains `timestamp` field that gives any extra information over just the plain `chargingState` (which is live visible on the platform). If the developer cannot put things like `url` there, then it is not very useful for both the user and the dev (except for local testing).

So we agreed (feel free to put your arguments in) that we should remove the current `chargingLogDataset` functionality before we figure out how to do it in a more bulletproof way. It shouldn't be any breaking adding it in later and it is not really required to make it work, devs can always to it themselves.
a) There are some ideas in the platform team about making this as platform feature that would not require storing it on user's account. Moving this forward depends of the feedback from both creators and users.
b) We might want to keep this functionality (ideally with the dev-provided metadata) for devs only to help with building the Actor. Now the question is how to do that in a way that doesn't escape to production.

In any way, I think we can ship this without the dataset for now so we have it out and think about the solution together. But happy to hear other arguments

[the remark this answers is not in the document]
metalwarrior665: @janbuchar @B4nan @netmilk Ok, we had a chat with @mhamas and there are several serious concerns with the debug dataset that didn't cross my mind previously.

Consider there can be use-cases where developer would charge per LLM token so each charge would be in thousands like Actor.charge(llm-token, 4562). This would create a huge pressure on the dataset, namely
a) Pushing thousands takes some time
b) push items cost for developer
c) timed storage cost for user
I didn't fully realizes that after stripping the developer provided metadata for each event, the chargingLogDataset now only contains timestamp field that gives any extra information over just the plain chargingState (which is live visible on the platform). If the developer cannot put things like url there, then it is not very useful for both the user and the dev (except for local testing).
So we agreed (feel free to put your arguments in) that we should remove the current chargingLogDataset functionality before we figure out how to do it in a more bulletproof way. It shouldn't be any breaking adding it in later and it is not really required to make it work, devs can always to it themselves.
a) There are some ideas in the platform team about making this as platform feature that would not require storing it on user's account. Moving this forward depends of the feedback from both creators and users.
b) We might want to keep this functionality (ideally with the dev-provided metadata) for devs only to help with building the Actor. Now the question is how to do that in a way that doesn't escape to production.

In any way, I think we can ship this without the dataset for now so we have it out and think about the solution together. But happy to hear other arguments

metalwarrior665 (changes requested):

B4nan:
> So we agreed (feel free to put your arguments in) that we should remove the current chargingLogDataset functionality before we figure out how to do it in a more bulletproof way. It shouldn't be any breaking adding it in later and it is not really required to make it work, devs can always to it themselves.

Or we could make it opt-in and mark it as experimental instead of removing it completely?

mhamas:
@B4nan I don't think it should be even opt-in, because it directly harms the end-user running the Actor by creating an additional dataset, for which timed storage they have to pay. Plus they have no idea why the dataset is there, and that it even exists. So I'd say let's not have it available in production at all in the first version.

B4nan:
Well, if it's opt-in, they need to enable it, so they should have at least _some_ idea. 

But no strong opinions here.

packages/apify/src/internals/charging.ts
janbuchar: The value depends on an environment variable set by the platform and it is used to distinguish local runs from runs on Apify. It's ubiquitous in the SDK code, so I don't see the need to explain it here.

packages/apify/src/internals/charging.ts
janbuchar: This was taken verbatim from the PoC made by @metalwarrior665 - I guess he can provide a better answer than me.

janbuchar:
If @B4nan means opt-in as in "user has to toggle a checkbox in Actor input", that would be safe, provided that the Creator knows what they're doing.

Otherwise it will incur unexpected costs for the end user if the Creator enables it.

packages/apify/src/internals/charging.ts
B4nan: Note that we also have `Actor.isAtHome()`, so this shouldn't be anything new even for our users, not just for contributors.

https://docs.apify.com/sdk/js/reference/class/Actor#isAtHome

B4nan:
Huh, what else can opt-in mean?

packages/apify/src/internals/charging.ts
janbuchar: Changed to a type exportd by the jS client.

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
B4nan: because of the `!` at the end, but I already pointed out how this could be refactored so its not needed.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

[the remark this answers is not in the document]
janbuchar: Also, typescript doesn't know that if the Actor is running of Apify, the run_id will be present in the configuration. I agree that it's safer to make sure.

janbuchar:
It can mean "End user can toggle it" and "Actor developer can toggle it", and those have different implications.

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
janbuchar: After some consideration, I'm in favor of making this opt-in on the code level, probably with some experimental flag. There are hundreds of other ways a Creator may waste money of their users, so I wouldn't stress over this to the point of removing this functionality entirely.

If the platform team is interested in supporting this on a higher level, all they would have to do would be to provide a charging dataset ID via environment variables.

Also, did you consider not making a separate dataset row for each occurence of an event when somebody does `Actor.charge('asdf', 33333)`? It doesn't fix everything, but it's something.

mhamas:
> In that case, we could print a warning when this is enabled, saying this shouldn't be used in production since it can cause additional price bumps and/or performance issues.
We can print a warning, or we can just remove it to be super sure somebody doesn't check it in by mistake. There is currently really no reason why this should be ever enabled on production for end users.

If you really want to keep it in as general opt-in, is it possible to know in SDK "I'm running in production and I'm a public Actor" or "I'm running in production and I'm not being run by the owner of the Actor" and then throw from the SDK at that point to stop the Actor?

B4nan:
> We can print a warning, or we can just remove it to be super sure somebody doesn't check it in by mistake :-).

You keep talking about a checkbox, I am talking about an option in `Actor` config, this is SDK, we are not creating any input schema here. So an explicit line in code doing something like `Configuration.set('enableChargingDataset', true)`.

> There is currently really no reason why this should be ever enabled on production for end users.

That's quite a weird reason to remove some functionality, you don't just run code in production, you need to first develop it. Wasn't this always meant for development? 

If you don't see a value during development, that would be another story.

metalwarrior665:
Btw recently we added periodic dump of chargingState to KV store locally. That feels like more conscise solution than bunch of individual dataset items files if it is only for local dev and we are not adding any other context data for each event.

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,281 @@
...
+                })();
+
+                this.chargingLogDataset = await Dataset.open(datasetId);
+            } else {
+                if (this.purgeChargingLogDataset) {
+                    const dataset = await Dataset.open(this.LOCAL_CHARGING_LOG_DATASET_NAME);
+                    await dataset.drop();
+                }
+
+                this.chargingLogDataset = await Dataset.open(this.LOCAL_CHARGING_LOG_DATASET_NAME);
```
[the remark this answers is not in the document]
janbuchar: Looks like it does, but I'll try to prove that with a test

packages/apify/src/internals/charging.ts
```diff
@@ -253,15 +265,17 @@ export class ChargingManager {
             throw new Error('ChargingManager is not initialized');
         }
 
-        if (!this.pricingInfo[eventName]) {
+        const price = this.isAtHome ? this.pricingInfo[eventName].price : 1; // Use a nonzero price for local development so that the maximum budget can be reached
```
metalwarrior665, line 268: This feels weird and a step down from our reference implementation. We allowed passing in run ID so you can use real pricing for that project (I guess that would still work if we pass both `ACTOR_RUN_ID` and `IS_AT_HOME`), here we hardcode it at $1. I guess you would be ok to pass in the pricing somehow but we just didn't figure how?

B4nan: I really didn't like the idea that you first need to publish the actor to figure things out, we need to do better. What about having default pricing info in the configuration too. Would be used locally only, and we could figure out a way to set it up on platform if its not there (and override it with whatever is set up on the platform otherwise).

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
mhamas: > If the platform team is interested in supporting this on a higher level, all they would have to do would be to provide a charging dataset ID via environment variables.

If we were to do this in the future, we would do it automatically in the API, it'd be on for all PPE runs and hence no special dataset ID would be passed to the Actor. Actor wouldn't be even aware that this is happening behind the scenes.

packages/apify/src/internals/charging.ts
janbuchar: It's a hard pass on copying the pricing from a platform run from me as well - it's just too involved and convoluted. But I can imagine that any alternative I'd implement would end up in endless bikeshedding, so I'd prefer to resolve this in a separate PR.

metalwarrior665: Ok, lets resolve later, not a huge deal and there is workaround. Config file is good but Im afraid of support questions why it didnt propagate to platform.

B4nan: We could have an info log when we use the platform pricing and config has one. It would be right at the top, so quite easy to see.

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
janbuchar: Authoritative as it sounds, I don't think that's the only viable solution. Bear in mind that this would not help at all with local develoment, and we'd probably have to keep the current implementation of the charging log dataset anyways...

packages/apify/src/internals/charging.ts
metalwarrior665: And I took it verbatim from someone else :D Please don't take me as expert in floats and do your own research. But this at least fixes the issue with the sum
```
(0.1 + 0.2) // => // 0.30000000000000004
```

```
Number((0.1 + 0.2).toFixed(6))  // => 0.3
```

This then makes the problem when calculating how many events we can still do. As this exact problem happened on the platform recently, we should have unified solution for it

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,307 @@
...
+                await this.apifyClient.run(this.actorRunId!).charge({ eventName, count: chargedCount });
+            } else {
+                log.warning(`Attempting to charge for an unknown event '${eventName}'`);
+            }
+        }
+
+        const timestamp = new Date().toISOString();
+
+        if (this.chargingLogDataset !== undefined) {
+            for (let i = 0; i < chargedCount; i++) {
```
mhamas, line 204: Suggestion: why not push one item for the whole `chargedCount`? If the developer wants to simultaneously charge N things, I would expect a single item would suffice for debugging. Charging line by line will potentially create a lot of data with no real value added. Also, doing it like this rapidly increases the potential cost incurred by using this (by mistake) in production.

mhamas (approved):
The PPE related matters look good to me, left some final thoughts on the log dataset.

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,305 @@
...
+            const datasetId = await this.ensureChargingLogDatasetOnPlatform();
+
+            this.chargingLogDataset = await Dataset.open(datasetId);
+        } else {
+            if (this.purgeChargingLogDataset) {
+                const dataset = await Dataset.open(this.LOCAL_CHARGING_LOG_DATASET_NAME);
+                await dataset.drop();
+            }
+
+            this.chargingLogDataset = await Dataset.open(this.LOCAL_CHARGING_LOG_DATASET_NAME);
```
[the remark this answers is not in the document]
tugkan: > I'll just state one more time that I think that we really want to make sure that the log dataset will **never** be used in production for Actors deployed to the store. Only for the testing by the developer. There is currently 0 value to use it for the end users, it'll only bring them harm, confusion, and potentially huge costs.
> 
> I'd be more conservative here and don't even allow the creation of the dataset in the cloud, only locally. I'll not block this PR on it, it's ultimately up to you and was already discussed a lot, but please let's make sure the developers are really aware of this, and let's minimize the risk of this them getting this wrong.
> 
> Also, if anytime in the future, we have a reason to start using this dataset for end users, let's discuss this internally as that functionality would then probably belong to the platform itself, not SDK. Thanks!

I think this is the `METADATA_DATASET` in @metalwarrior665 repository. If so, that specific implementation did not provide any value to our tests except for confusion. It doesn't get purged properly between runs (locally), and we do not utilize it. The end users will see thousands of this dataset and they will ask the reason for the customer service.

However, I can understand the main intention for this specific implementation. I believe users would like to see a detailed cost report and see who paid what with which amount. If this can be shown differently in the platform, we can only use this on the development as far as it gets purged.

metalwarrior665 (approved):

packages/apify/src/actor.ts
```diff
@@ -896,6 +970,25 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+     * @param options The name of the event to charge for and the number of events to be charged.
+     * @ignore
+     */
+    async charge(options: ChargeOptions): Promise<ChargeResult> {
+        this._ensureActorInit('charge');
+        return this.chargingManager.charge(options);
+    }
+
+    /**
+     * Retrieve the charging manager to access granular pricing information.
```
B4nan, line 985:
```suggestion
     * Retrieve the charging manager to access granular pricing information.
     * @ignore
```

packages/apify/src/internals/charging.ts
janbuchar: This is taken from the PoC by @metalwarrior665, but I agree with your suggestion. It's small enough of a change to add that.

packages/apify/src/internals/charging.ts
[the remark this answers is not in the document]
janbuchar: Sure, I can disable it in the cloud. This implementation should support purging locally.

B4nan (approved):

package.json
```diff
@@ -40,7 +40,10 @@
         "build": "turbo run build",
         "ci:build": "turbo run build --cache-dir=\".turbo\"",
         "test": "vitest run --silent",
-        "test:e2e": "node test/e2e/run.mjs",
+        "test:e2e": "npm run test:e2e:scrapers && npm run test:e2e:sdk",
+        "test:e2e:scrapers": "node test/e2e/runScraperTests.mjs",
+        "test:e2e:sdk": "npm run test:e2e:sdk:tarball && node test/e2e/runSdkTests.mjs",
+        "test:e2e:sdk:tarball": "npm run build && cd packages/apify && mv $(npm pack | tail -n 1) ../../test/e2e/apify.tgz",
```
vladfrangu, line 46: not very portable but c'est la vie

packages/apify/src/internals/charging.ts
```diff
@@ -0,0 +1,310 @@
...
+import { Configuration } from '../configuration.js';
+
+/**
+ * Handles pay-per-event charging.
+ *
+ * @internal
+ */
+export class ChargingManager {
+    readonly LOCAL_CHARGING_LOG_DATASET_NAME = 'charging_log';
+    readonly PLATFORM_CHARGING_LOG_DATASET_ID_KEY = 'CHARGING_LOG_DATASET_ID';
```
vladfrangu, line 14: This feels like it should come from apify/consts

vladfrangu (approved):

packages/apify/src/internals/charging.ts
metalwarrior665: The idea was that there will be mapping of items and events and if you push a list, each event might have different metadata. But we diverged from the idea anyway so feel free to change

packages/apify/src/internals/charging.ts
janbuchar: Yeah, I admit I have deep disdain for that package. But I honestly believe that it is justified not to use it here since this is all very adhoc and may get removed. Might be a reason to make these private, huh...

package.json
janbuchar: If it works in CI and on my laptop, it's good enough for me.

packages/apify/src/actor.ts
metalwarrior665: FYI: By moving the push before the charge, we introduced a small race condition where as we are close to the max budget, more pushes can happen concurrently and we would push a few more items than the limit. I think this might happen maybe in 1/1000 runs, maybe even less (depends on the size of the data push which determines the latency of it).

I thought a bit about fixes for this and haven't come up with anything elegant. You could have a new `bookFutureCharge` method to pre-register events but it would introduce a new complex API. Or we could patch it on the `Actor` class bypassing the manager and just handle in the `pushData`. All of it introduces extra complexities. So for now I'm parking that thought because I don't think it is a big deal.

PR #356, opened 2025-02-10
docs: Actor.charge guide

### LINKED ISSUE #355: Add a docs page about Actor.charge
https://apify.slack.com/archives/C084Y2Q73B2/p1738915786185369

janbuchar:
- closes #355


docs/guides/code/actor_charge.ts
```diff
@@ -0,0 +1,32 @@
...
+await Actor.charge({ eventName: 'init' });
+// highlight-end
+
+// Prepare some mock results
+const result = [
+    { word: "Lorem" }, 
+    { word: "Ipsum" }, 
+    { word: "Dolor" }, 
+    { word: "Sit" }, 
+    { word: "Amet" },
```
B4nan, line 16: nit: use single quotes

docs/guides/pay_per_event.mdx
```diff
@@ -0,0 +1,37 @@
+---
+id: pay-per-event
```
B4nan, line 2: since we shipped this as v3.3, we should update the v3.3 docs snapshot too

(but lets keep that as the final step before merging for easier review)

packages/apify/src/internals/charging.ts
```diff
@@ -6,8 +6,6 @@ import { Configuration } from '../configuration.js';
 
 /**
  * Handles pay-per-event charging.
- *
- * @internal
```
B4nan, line 10: we'll need to regenerate the typedoc json for this to appear in the stable docs too

B4nan (changes requested):

docs/guides/pay_per_event.mdx
janbuchar: How does one go about that?

B4nan: for API docs: inside the website folder just run `npx docusaurus api:version 3.3`, maybe you'll need to remove the JSON file first, but I don't think its necessary

the regular docs are done the same way via `npx docusaurus docs:version 3.3`, but that generates a lot of files and touches the versions.json, so you would have to first wipe all of that. 

---

anyway, i just found out i forgot to do this, and we are missing the v3.3 snapshot completely :D so once we merge this, it's as simple as calling those and committing what it generates

```
cd website
npx docusaurus docs:version 3.3
npx docusaurus api:version 3.3
```

B4nan: also we'll need to reindex algolia, so ping me once its done

(another thing we should ideally automate)

metalwarrior665 (approved):

docs/guides/pay_per_event.mdx
```diff
@@ -0,0 +1,37 @@
...
+Then you just push your code to Apify and that's it! The SDK will even keep track of the max total charge setting for you, so you will not provide more value than what the user chose to pay for.
+
+If you need finer control over charging, you can access call <ApiLink to="class/Actor#getChargingManager">`Actor.getChargingManager()`</ApiLink> to access the <ApiLink to="class/ChargingManager">`ChargingManager`</ApiLink>, which can provide more detailed information - for example how many events of each type can be charged before reaching the configured limit.
+
+## Local development
+
+It is encouraged to test your monetization code on your machine before releasing it to the public. To tell your Actor that it should work in pay-per-event mode, pass it the `ACTOR_TEST_PAY_PER_EVENT` environment variable:
+
+```shell
+ACTOR_TEST_PAY_PER_EVENT=true npm start
```
grudiger, line 32: Nice

mhamas (changes requested):
🚀 thanks!

Is there any advice on how to use SDK when migrating Actor from PPR to PPE? Before the switch happens, the SDK needs to be used to start calling `Actor.charge()` in the code, but these calls must not reach the API as it would fail since Actor is not yet PPE? How to best handle this case with SDK? Can we add this to the guide? Thanks!

metalwarrior665:
We made this huge mistake of not putting PPR into SDK years ago so it will bite us back a few more times instead of working seamlessly :) But they can do simply

```js
if (Actor.getChargingManager().getPricingInfo().isPayPerEvent) {
    await Actor.pushData(data, 'my-event')
} else {
    await pushDataPPRMaxAware(data)
}
```

B4nan:
I thought we already discussed this somewhere, IMO actor code should be either PPR or PPE, not both. Different versions are different builds, the switch on the platform should be bound to the version change (a tag would make the most sense here to allow future updates of both variants).

metalwarrior665:
That's a decision developers should make. I think most (like us internally) will opt for having a single version for these reasons:
1. You must have single pricing for your Actor so version bump doesn't matter because the old code will not work anyway. Unless you want to explicitly error out for users who don't migrate to the new version.
2. The pricing change happens by platform at "sometime at given day". So instead of waiting to rebuild, you just make it work automatically. This is also nicer if you want to have the PPE code ready but still work on that codebase, no need to backport fixes etc.

I'm sure there will be Actors or devs where it will make sense to force a version switch. But I would definitely not make that mandatory or default

grudiger:
> That's a decision developers should make. I think most (like us internally) will opt for having a single version for these reasons:
> 
>     1. You must have single pricing for your Actor so version bump doesn't matter because the old code will not work anyway. Unless you want to explicitly error out for users who don't migrate to the new version.
> 
>     2. The pricing change happens by platform at "sometime at given day". So instead of waiting to rebuild, you just make it work automatically. This is also nicer if you want to have the PPE code ready weeks ahead for testing but still work on that codebase, no need to backport fixes etc.
> 
> 
> I'm sure there will be Actors or devs where it will make sense to force a version switch. But I would definitely not make that mandatory or default. The PPR code is really a few lines of code that doesn't add almost any complexity so there is not much pressure to fully migrate out of it.

Agree

janbuchar:
@mhamas I added a section about transitioning to price-per-event, please check it out

docs/guides/code/conditional_actor_charge.ts
```diff
@@ -0,0 +1,18 @@
+import { Actor } from 'apify';
+
+await Actor.init();
+
+let chargedItems = 0;
```
metalwarrior665, line 5:
```suggestion
// Check the dataset because there might already be items if the run migrated or was restarted
const defaultDataset = Actor.openDataset();
let chargedItems = (await defaultDataset.getInfo()!).itemCount;
```

metalwarrior665 (changes requested):

mhamas (approved):
Thanks

B4nan (approved):

metalwarrior665 (approved):
You can merg afterwards

docs/guides/code/conditional_actor_charge.ts
```diff
@@ -0,0 +1,20 @@
+import { Actor } from 'apify';
+
+await Actor.init();
+
+// Check the dataset because there might already be items if the run migrated or was restarted
+const defaultDataset = Actor.openDataset();
+let chargedItems = (await defaultDataset.getInfo()!).itemCount;
```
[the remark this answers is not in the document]
janbuchar: No worries, it happens to all of us :grin:

PR #358, opened 2025-02-17
feat: sign record's public url

danpoletaev:
This PR is part of [Issue #19363](https://github.com/apify/apify-core/issues/19363), which updates `KeyValueStore.getPublicUrl(recordKey)` to generate signed links using HMAC.

**This PR**:
- creates signature and appends it to the public URL of the record in KV store

P.S. Before merging, we need to wait for the release of Crawlee, so we can update version of Crawlee.

Previous PR that adds storageObject to Crawlee is [here](https://github.com/apify/crawlee/pull/2837)


packages/apify/src/key_value_store.ts
```diff
@@ -19,7 +20,11 @@ export class KeyValueStore extends CoreKeyValueStore {
             return getPublicUrl.call(this, key);
         }
 
-        return `https://api.apify.com/v2/key-value-stores/${this.id}/records/${key}`;
+        if (!this.storageObject?.urlSigningSecretKea) {
+            return `https://api.apify.com/v2/key-value-stores/${this.id}/records/${key}`;
+        }
+
+        return `https://api.apify.com/v2/key-value-stores/${this.id}/records/${key}?signature=${createHmacSignature(this.storageObject.urlSigningSecretKea as string, key)}`;
```
barjin, line 27: Can we pls compose the URL with `URL.searchParams` and similar? Maybe concatting the path with `path.posix.join()` too?

barjin, line 27: Can we pls compose the URL with `URL.searchParams` and similar? Maybe concatting the path with `path.posix.join()` too?

packages/apify/src/key_value_store.ts
```diff
@@ -19,7 +20,11 @@ export class KeyValueStore extends CoreKeyValueStore {
             return getPublicUrl.call(this, key);
         }
 
-        return `https://api.apify.com/v2/key-value-stores/${this.id}/records/${key}`;
+        if (!this.storageObject?.urlSigningSecretKea) {
```
barjin, line 23: `urlSigningSecretKea` instead of `...key` might be a typo?

packages/apify/src/key_value_store.ts
barjin: Aside from ensuring the correct URL syntax, this procedural approach might also allow us to only specify the pathname once (and just conditionally add the query parameter if the secret exists).

barjin (approved):
LGTM, thanks! 

(as long as we wait for the new stable Crawlee version, but I see you mentioned that in the PR description already 👍🏽 )

packages/apify/src/key_value_store.ts
```diff
@@ -19,7 +20,13 @@ export class KeyValueStore extends CoreKeyValueStore {
             return getPublicUrl.call(this, key);
         }
 
-        return `https://api.apify.com/v2/key-value-stores/${this.id}/records/${key}`;
+        const publicUrl = new URL(`https://api.apify.com/v2/key-value-stores/${this.id}/records/${key}`);
```
tobice, line 23: (nit) Isn't the API base URL available in the env variables?

That way the Actor will create valid urls in all environments. You could leverage that to create an actual e2e test that not only creates & runs an Actor that produces the signed URL, but actually tests the URL against Apify API.

(The current test below basically just copies the implementation... you are calling the same code twice and checking that it produces the same value :sweat_smile: )

tobice (changes requested):

packages/apify/src/configuration.ts
```diff
@@ -183,6 +185,7 @@ export class Configuration extends CoreConfiguration {
         defaultRequestQueueId: LOCAL_ACTOR_ENV_VARS[ACTOR_ENV_VARS.DEFAULT_REQUEST_QUEUE_ID],
         inputKey: 'INPUT',
         apiBaseUrl: 'https://api.apify.com',
+        apiPublicBaseUrl: 'https://api.apify.com',
```
B4nan, line 188: why another option for this?

package.json
```diff
@@ -67,7 +67,7 @@
         "@typescript-eslint/eslint-plugin": "^7.0.0",
         "@typescript-eslint/parser": "^7.0.0",
         "commitlint": "^19.3.0",
-        "crawlee": "^3.11.5",
+        "crawlee": "^3.13",
```
B4nan, line 70: please keep the full version definition here

packages/apify/package.json
```diff
@@ -58,15 +58,15 @@
...
         "@apify/timeout": "^0.3.0",
-        "@apify/utilities": "^2.9.3",
-        "@crawlee/core": "^3.9.0",
-        "@crawlee/types": "^3.9.0",
-        "@crawlee/utils": "^3.9.0",
-        "apify-client": "^2.12.0",
+        "@apify/utilities": "^2.13.0",
+        "@crawlee/core": "^3.12.3-beta.12",
+        "@crawlee/types": "^3.12.3-beta.12",
+        "@crawlee/utils": "^3.12.3-beta.12",
```
B4nan, line 64: and don't forget to bump this

packages/apify/package.json
```diff
@@ -58,15 +58,15 @@
...
+        "apify-client": "^2.11.1",
```
B4nan, line 65: why the downgrade here? 👀

packages/apify/src/key_value_store.ts
```diff
@@ -15,11 +16,18 @@ export class KeyValueStore extends CoreKeyValueStore {
...
      */
     override getPublicUrl(key: string): string {
-        if (!(this.config as Configuration).get('isAtHome') && getPublicUrl) {
+        const config = this.config as Configuration;
+        if (!config.get('isAtHome') && getPublicUrl) {
             return getPublicUrl.call(this, key);
         }
 
-        return `https://api.apify.com/v2/key-value-stores/${this.id}/records/${key}`;
+        const publicUrl = new URL(`${config.get('apiPublicBaseUrl', 'https://api.apify.com')}/v2/key-value-stores/${this.id}/records/${key}`);
```
B4nan, line 24: no need for the explicit default here, its the exact same value as you define on the config level

B4nan (changes requested):

packages/apify/src/key_value_store.ts
B4nan: i _think_ there was a reason for this. if you ask what reason, i have no idea, but there is something at the back of my mind telling me there was one 🙃

test/e2e/sdk/publicUrl/src/main.mjs
```diff
@@ -0,0 +1,21 @@
+import { Actor, log } from 'apify';
+
+// Also needs to be changed in test.mjs
+const PUBLIC_RECORD_KEY = 'public-record-key';
```
tobice, line 4: (opt) How about you pass this as input to the Actor? All three values...

- data to upload
- key where to upload the data
- key where to store the public URL

Then you don't need to store anything in two places and the test will be more readable (not that it is super complicated or anything :smile: )

tobice (changes requested):
Same as @B4nan  wondering why `apiPublicBaseUrl` when we already have `apiBaseUrl`.

As in, I know that we're injecting as env variable `APIFY_API_PUBLIC_BASE_URL`... but I had no idea we have apparently also `APIFY_API_BASE_URL` :sweat_smile:

B4nan:
> As in, I know that we're injecting as env variable APIFY_API_PUBLIC_BASE_URL... but I had no idea we have apparently also APIFY_API_BASE_URL 😅

Hmm, maybe it was about that actually. `APIFY_API_BASE_URL` is some local URL that won't work from outside of the platform/worker, right?

tobice:
> Hmm, maybe it was about that actually. APIFY_API_BASE_URL is some local URL that won't work from outside of the platform/worker, right?

I checked the code, and that exactly seems to be the case :+1: I think the traffic from Actors to Apify API does not go through the Internet. It's routed internally through AWS, presumably as it's cheaper.

So `APIFY_API_PUBLIC_BASE_URL` is indeed the correct value to use in this particular case.

danpoletaev:
This is right, that's why I used `APIFY_API_PUBLIC_BASE_URL`, because `apiBaseUrl` was an IP address

packages/apify/src/key_value_store.ts
danpoletaev: @B4nan I think it should be safe to use `config.get('apiPublicBaseUrl')` variable as it has fallback to `https://api.apify.com`?

packages/apify/src/configuration.ts
danpoletaev: explanation below

test/e2e/sdk/publicUrl/src/main.mjs
danpoletaev: That's a good point, thank you 🙏  Refactored it

packages/apify/src/configuration.ts
```diff
@@ -16,6 +16,7 @@ export interface ConfigurationOptions extends CoreConfigurationOptions {
     actorRunId?: string;
     actorTaskId?: string;
     apiBaseUrl?: string;
+    apiPublicBaseUrl?: string;
```
B4nan, line 19: lets add a comment here about the difference between `apiBaseUrl` and `apiPublicBaseUrl`

B4nan (approved):

test/e2e/sdk/publicUrl/test.mjs
```diff
@@ -0,0 +1,19 @@
...
+
+const PUBLIC_DATA = { exposedData: 'test' };
+
+const client = new ApifyClient({
+    token: process.env.APIFY_TOKEN,
+});
+
+const actor = client.actor(process.argv[2]);
+
+const run = await actor.call({ data: PUBLIC_DATA, recordKey: 'public-record-key' }, { waitSecs: 15 });
```
tobice, line 13: (opt) `public-record-key` is actually not needed to be passed as input, it's completely internal :see_no_evil: You will get it as part of the public URL, but you don't care what's in it, you only care that the URL works.

 What you might want to pass in is `urlToPublicData`. But you could also just upload the URL to the dataset and fetch it from there.

Anyway, sorry for all the back and forths on this :sweat_smile:

tobice (approved):

PR #359, opened 2025-02-17
docs: Snapshot version 3.3

janbuchar:


B4nan (approved):

B4nan:
hmm the broken build is weird, i recall we were dealing with this somewhere already

barjin:
It's the broken docs snapshotting in `@apify/docusaurus-plugin-typedoc-api`. You need to generate the snapshot using an older version of the package. The same thing has happened in Crawlee already - while we know how to circumvent it, I have no idea what causes it (yet) - both versions generate similar-enough JSONs. I can generate the snapshot for you in a sec

janbuchar:
That would be nice, thank you very much @barjin!

barjin:
obrazek

Sorry, something came up - regenerating the `typedoc.json` with `@apify/docusaurus-plugin-typedoc-api@3.0.1` fixes the issue.

My commits also added the recently merged guides (#356 ) to the `3.3` versioned docs, so they show up for users that are not on `next`. If this is not the desired behaviour, feel free to remove the files (or let me know :))

PR #361, opened 2025-02-20
fix: ensure `maxTotalChargeUsd` is correctly mapped to number, consider empty string as infinity

B4nan:
The `ACTOR_MAX_TOTAL_CHARGE_USD` env var wasnt properly marked as integer, and the fallback to infinity wasn't working in case the value was an empty string, since `??` fallbacks only for null/undefined values.

### DIFF
--- a/test/apify/actor.test.ts
+++ b/test/apify/actor.test.ts
@@ -612,6 +612,7 @@ describe('Actor', () => {
                 expect(openStorageSpy).toBeCalledWith(queueId, sdk.apifyClient);
                 expect(openStorageSpy).toBeCalledTimes(1);
 
+                // @ts-expect-error private prop
                 expect(queue.initialCount).toBe(10);
             });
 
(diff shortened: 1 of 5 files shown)


test/apify/actor.test.ts
```diff
@@ -1030,4 +1031,18 @@ describe('Actor', () => {
             expect(pushDataSpy).toHaveBeenCalledWith({ hello: 'apify' });
         });
     });
+
+    describe('Actor.config and PPE', () => {
+        test('should work', async () => {
+            await Actor.init();
+            process.env.ACTOR_MAX_TOTAL_CHARGE_USD = '';
```
janbuchar, line 1038: What about this case: `process.env.ACTOR_MAX_TOTAL_CHARGE_USD = '0';`? In that case we probably don't want to coerce to `Infinity`...

B4nan: @mhamas thoughts? it sounds fine to me, `0` is often considered to be infinity.

if we want to distinguish empty string and zero, we'd have to adjust the logic in the config class a bit (and that would first need to happen in crawlee, but we can hotfix it from the SDK too now so we are not blocked by the crawlee release). i guess it would be fine if we just check for the empty string and map the value to `undefined`.

mhamas: 0 is not a valid value for the total max. I think coercing to infinity is fine in that case.

mhamas (approved):

PR #364, opened 2025-02-26
feat: add Camoufox-based generic Actor

barjin:
Adds a Playwright Scraper port that uses Camoufox as the browser backend [via Camoufox JS](https://github.com/apify/camoufox-js/)).

This PR removes the original `useChrome` and `browserLauncher` options from the new input schema (as the only available browser is now Camoufox) and adds a `Camoufox options` input section, which maps to the [options that can be passed to Camoufox](https://github.com/apify/camoufox-js/blob/286debe0171ad2b1870ac23869b68b1e4f16e47a/src/utils.ts#L300).

### DIFF
--- a/packages/actor-scraper/camoufox-scraper/README.md
+++ b/packages/actor-scraper/camoufox-scraper/README.md
@@ -1,0 +1,5 @@
+# Camoufox Scraper
+
+This is a port of [Playwright Scraper](https://apify.com/apify/playwright-scraper) that's using [Camoufox](https://camoufox.com/) instead of default Playwright Chromium.
+
+All input options are the same as in Playwright Scraper.
--- a/packages/actor-scraper/camoufox-scraper/README.md
+++ b/packages/actor-scraper/camoufox-scraper/README.md
@@ -18,7 +18,7 @@ To get started with Camoufox Scraper, you only need a few things. First, with `S
 
 The scraper starts by loading pages specified in the [**Start URLs**](#start-urls) input setting. You can make the scraper follow page links on the fly by setting a **[Link selector](#link-selector)**, **[Glob Patterns](#glob-patterns)** and/or **[Pseudo-URLs](#pseudo-urls)** to tell the scraper which links it should add to the crawler's request queue. This is useful for the recursive crawling of entire websites (e.g. finding all products available in an online store).
 
-To tell the scraper how to handle requests and extra data, you need to provide a **[Page function](#page-function)**, and optionally arrays of **[Pre-navigation hooks](#pre-navigation-hooks)** and **[Post-navigation hooks](#post-navigation-hooks)**. This is JavaScript code that is executed in the Node.js environment. Since the scraper uses the full-featured Chromium or Firefox browser, client-side logic to be executed within the context of the web-page can be done using the **[`page`](#page)** object within the Page function's context.
+To tell the scraper how to handle requests and extra data, you need to provide a **[Page function](#page-function)**, and optionally arrays of **[Pre-navigation hooks](#pre-navigation-hooks)** and **[Post-navigation hooks](#post-navigation-hooks)**. This is JavaScript code that is executed in the Node.js environment. Since the scraper uses the full-featured Camoufox browser, client-side logic to be executed within the context of the web-page can be done using the **[`page`](#page)** object within the Page function's context.
 
 In summary, Camoufox Scraper works as follows:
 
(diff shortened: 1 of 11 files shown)


barjin:
I see this mostly as a niche tool and the "next step" we can send users with antibot issues to. Also, a simple way of testing whether switching to Camoufox will help you (before actually modifying your project). 

On the other hand, it's yet another thing to maintain. Do you guys see any opportunity to share the codebase with Playwright Scraper? Or would it be too much work for too little gain in your eyes?

barjin:
One more thing - @Pijukatel already released [Camoufox Scraper](https://apify.com/josef.prochazka/camoufox-scraper) on Apify (great job btw). It's basically the same as in this PR, only for Python (Page function has to be in Python).

Should we name this one Camoufox Scraper JS (as users would expect Camoufox to be Python), or should we switch the names (so this gets closer to the other generic Actor projects)?

Pijukatel:
> One more thing - @Pijukatel already released [Camoufox Scraper](https://apify.com/josef.prochazka/camoufox-scraper) on Apify (great job btw). It's basically the same as in this PR, only for Python (Page function has to be in Python).
> 
> Should we name this one Camoufox Scraper JS (as users would expect Camoufox to be Python), or should we switch the names (so this gets closer to the other generic Actor projects)?

No preference here. I checked stats about Actor usage and no one seems to be using it so far, so renaming should not be problem at all.

I see you exposed many configurations options, this is definitely good improvement.

packages/actor-scraper/camoufox-scraper/README.md
```diff
@@ -0,0 +1,645 @@
...
+
+The scraper starts by loading pages specified in the [**Start URLs**](#start-urls) input setting. You can make the scraper follow page links on the fly by setting a **[Link selector](#link-selector)**, **[Glob Patterns](#glob-patterns)** and/or **[Pseudo-URLs](#pseudo-urls)** to tell the scraper which links it should add to the crawler's request queue. This is useful for the recursive crawling of entire websites (e.g. finding all products available in an online store).
+
+To tell the scraper how to handle requests and extra data, you need to provide a **[Page function](#page-function)**, and optionally arrays of **[Pre-navigation hooks](#pre-navigation-hooks)** and **[Post-navigation hooks](#post-navigation-hooks)**. This is JavaScript code that is executed in the Node.js environment. Since the scraper uses the full-featured Chromium or Firefox browser, client-side logic to be executed within the context of the web-page can be done using the **[`page`](#page)** object within the Page function's context.
```
Pijukatel, line 21: >  Since the scraper uses the full-featured Chromium or Firefox browser

This does not seem be updated for the Camoufox case

packages/actor-scraper/camoufox-scraper/README.md
Pijukatel, line 1: Maybe a short comment about drawbacks. Camoufox seems to be way slower than Playwright+Firefox and based on my several runs it seems to require way more memory to run.

B4nan:
No strong preference about the naming, I guess I would keep the existing one as is, or maybe add `for Python` to its name, but not to the identifier (or how to call it, I mean `josef.prochazka/camoufox-scraper` can stay). For one for JS could also state it in the name, but the identifier would be just `apify/camoufox-scraper`. Or we can add the language there as well, but then I would like to see it in both too.

B4nan (approved):
(only had a brief look)

---

re code duplication, i'd keep it as is for now, unless the code from the playwright scraper can be reused without any modifications easily.

packages/actor-scraper/camoufox-scraper/src/internals/consts.ts
```diff
@@ -0,0 +1,49 @@
+import { Session, ProxyConfigurationOptions, GlobInput, RegExpInput, PseudoUrlInput, RequestOptions } from '@crawlee/core';
+import { Dictionary } from '@crawlee/utils';
+
+/**
+ * Replicates the INPUT_SCHEMA with TypeScript types for quick reference
+ * and IDE type check integration.
+ */
+export interface Input {
```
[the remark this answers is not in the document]
barjin: Good point, I just added them, you cannot be too safe 👍🏽

barjin:
> unless the code from the playwright scraper can be reused without any modifications 

those two Actors are basically identical, the only change in code is

```diff
new PlaywrightCrawler({
+    launcher: firefox,
     launchContext: {
+       launchOptions: await camoufoxLaunchOptions({ ...input })
     }
})
```

> can be reused without any modifications **easily**

That's the problem, the only idea I have is a layered Docker build, but that would likely mess up the Actor build on Platform anyway. Let's keep it as it I suppose, we can always polish later.

Pijukatel (approved):

PR #379, opened 2025-04-22
fix: don't process `null` / `undefined` pageFunction results

### LINKED ISSUE #353: (web-scraper): `Max result records` counts processed pages, not results
As reported in a [user issue](https://console.apify.com/actors/moJRLRc85AitArpNN/issues/dhWm8huuzAnsTpqcz), returning `undefined` from the Page Function still fires a request to the Dataset API (adding no item) and counts towards the `Max result records` (`maxResultsPerCrawl`) crawler limit.

With the limit set to e.g. `10` and `pageFunction` returning `undefined` on every other request, the Actor will finish with 5 dataset items only (will stop prematurely, since the "limit has been hit"). 

This seems like a bug to me, so I'd fix it regardless of whether it's breaking. 

Reproduction scenario here: https://console.apify.com/view/runs/T3XtWRqfhCSFUeH7X

barjin:
Early returns from `_handleResult` on a nullish result. Saves Dataset API calls.

Closes #353.

### DIFF
--- a/packages/actor-scraper/camoufox-scraper/src/internals/crawler_setup.ts
+++ b/packages/actor-scraper/camoufox-scraper/src/internals/crawler_setup.ts
@@ -470,6 +470,8 @@ export class CrawlerSetup implements CrawlerSetupOptions {
         pageFunctionResult?: Dictionary,
         isError?: boolean,
     ) {
+        if (!pageFunctionResult) return;
+
         const payload = tools.createDatasetPayload(
             request,
             response,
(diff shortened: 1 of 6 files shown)


barjin:
Oops, on second look, this comment explains the reasoning behind the original design:

https://github.com/apify/apify-sdk-js/blob/49c666ebbfa382398a382ab42135cba71c949b4c/packages/scraper-tools/src/tools.ts#L135-L137

Even `null` / `undefined` pageFunction results generate dataset objects (with `#error` and `#debug` fields). This seems to be the expected behaviour (also with Apify API's `clean` query param, which hides the `#`-prefixed fields). 

Closing this PR and closing the original issue as `wontfix`.

PR #380, opened 2025-04-22
chore(ci): add automatic release workflow for generic Actors

barjin:
Adds a manual-dispatch CI workflow for running generic Actor builds on Apify.

### DIFF
--- a/.github/workflows/release-generic-actors.yaml
+++ b/.github/workflows/release-generic-actors.yaml
@@ -107,10 +107,10 @@ jobs:
               id: calculate-versions
               if: matrix.settings.should-build == 'true'
               run: |
-                  if [ "${{ github.event.inputs.build-channel }}" == "stable" ]; then
+                  if [ "${{ github.event.inputs.build-channel }}" = "stable" ]; then
                       echo "version=${{ matrix.settings.stable-version }}\n" >> $GITHUB_ENV
                       echo "build-tag=${{ matrix.settings.stable-build-tag }}\n" >> $GITHUB_ENV
-                  elif [ "${{ github.event.inputs.build-channel }}" == "development" ]; then
+                  elif [ "${{ github.event.inputs.build-channel }}" = "development" ]; then
                       echo "version=${{ matrix.settings.development-version }}\n" >> $GITHUB_ENV
                       echo "build-tag=${{ matrix.settings.development-build-tag }}\n" >> $GITHUB_ENV
                   else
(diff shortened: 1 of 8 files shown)


.github/workflows/release-generic-actors.yaml
```diff
@@ -0,0 +1,76 @@
...
+            camoufox-scraper:
+                description: apify/camoufox-scraper
+                type: boolean
+                required: false
+                default: true
+            version:
+                description: Version to build
+                type: string
+                required: true
+                default: '3.0'
```
B4nan, line 40: how does this work? some of the scrapers are on a lower version since they are newer, e.g. playwright and jsdom

also can this be used to trigger the development build of all of them at once?

barjin: > some of the scrapers are on a lower version since they are newer, e.g. playwright and jsdom

Ah, that's annoying, and I missed that. We can omit the version param altogether if we set the version right in the `.apify/actor.json` file. Right now, there's just a bunch of top-level `apify.json` files.

> also can this be used to trigger the development build of all of them at once?

I don't see why not, even though we'll need to make sure all the `development` versions match.

Thinking about it now, it might be better to replace the `build-tag` and `version` options with a "build type" `choice` (picker with `latest` / `beta` / `canary` or something similar) and have the logic in the action internally. We might want to match all the Actor versions on the platform, so the implementation is easier, but I'm not sure about the implications of this.

.github/workflows/release-generic-actors.yaml
```diff
@@ -52,25 +61,63 @@ jobs:
...
-              with:
-                  token: ${{ secrets.APIFY_ACTOR_BUILD_TOKEN }}
-                  build-tag: ${{ github.event.inputs.build-tag }}
-                  version: ${{ github.event.inputs.version }}
-                  working-directory: packages/actor-scraper/${{ matrix.settings.actor }}
+              run: |
+                  if [ "${{ github.event.inputs.build-channel }}" == "stable" ]; then
+                      echo "version=${{ matrix.settings.stable-version }}\n" >> $GITHUB_ENV
+                      echo "build-tag=${{ matrix.settings.stable-build-tag }}\n" >> $GITHUB_ENV
+                  elif [ "${{ github.event.inputs.build-channel }}" == "development" ]; then
```
janbuchar, line 113:
```suggestion
                  if [ "${{ github.event.inputs.build-channel }}" = "stable" ]; then
                      echo "version=${{ matrix.settings.stable-version }}\n" >> $GITHUB_ENV
                      echo "build-tag=${{ matrix.settings.stable-build-tag }}\n" >> $GITHUB_ENV
                  elif [ "${{ github.event.inputs.build-channel }}" = "development" ]; then
```

You're supposed to use `=` for string comparison, `==` is a non-standard extension :imp:

PR #385, opened 2025-05-05
fix: prefer proxy password from env var

stepskop:
- Fetch proxy password from `/user/me` using `APIFY_ENV_VARS.TOKEN` only when `APIFY_ENV_VARS.PROXY_PASSWORD` is not provided
- Proxy init no longer logs when there is a difference between the proxy password from the environmental variable and the one obtained from the API, since only one is going to be taken into consideration. When running on Apify, the passwords will always be different. 
- should solve the second part of: https://console.apify.com/actors/RB9HEZitC8hIUXAha/issues/exVPDR4pOsgqlkE5T

Closes [#20502](https://github.com/apify/apify-core/issues/20502)

### DIFF
--- a/packages/apify/src/proxy_configuration.ts
+++ b/packages/apify/src/proxy_configuration.ts
@@ -4,7 +4,6 @@ import type {
 } from '@crawlee/core';
 import { ProxyConfiguration as CoreProxyConfiguration } from '@crawlee/core';
 import { gotScraping } from '@crawlee/utils';
-import type { UserProxy } from 'apify-client';
 import ow from 'ow';
 
 import { APIFY_ENV_VARS, APIFY_PROXY_VALUE_REGEX } from '@apify/consts';
@@ -268,15 +267,15 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
                 if (Actor.isAtHome()) {
                     throw new Error(
                         `Apify Proxy password must be provided using options.password or the "${APIFY_ENV_VARS.PROXY_PASSWORD}" environment variable. ` +
-                            `If you add the "${APIFY_ENV_VARS.TOKEN}" environment variable, ` +
-                            `while not providing "${APIFY_ENV_VARS.PROXY_PASSWORD}", the password will be automatically inferred.`,
+                            `You can also provide your Apify token via the "${APIFY_ENV_VARS.TOKEN}" environment variable,` +
+                            `so that the SDK can fetch the proxy password from Apify API, when ${APIFY_ENV_VARS.PROXY_PASSWORD} is not defined`,
                     );
                 } else {
                     this.log.warning(
                         `No proxy password or token detected, running without proxy. To use Apify Proxy locally, ` +
                             `provide options.password or "${APIFY_ENV_VARS.PROXY_PASSWORD}" environment variable. ` +
-                            `If you add the "${APIFY_ENV_VARS.TOKEN}" environment variable, ` +
-                            `while not providing "${APIFY_ENV_VARS.PROXY_PASSWORD}", the password will be automatically inferred.`,
+                            `You can also provide your Apify token via the "${APIFY_ENV_VARS.TOKEN}" environment variable,` +
+                            `so that the SDK can fetch the proxy password from Apify API, when ${APIFY_ENV_VARS.PROXY_PASSWORD} is not defined`,
                     );
                 }
             }
@@ -435,33 +434,24 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
     }
 
     /**
-     * Checks if proxy password is provided in env, if not, fetches it from API using Apify Token
+     * Fetch & set the proxy password from Apify API if an Apify token is provided.
      */
     // TODO: Make this private
     protected async _setPasswordIfToken(): Promise<void> {
         const token = this.config.get('token');
 
-        if (token) {
-            let proxy: UserProxy;
-
-            try {
-                const user = await Actor.apifyClient.user().get();
-                proxy = user.proxy!;
-            } catch (error) {
-                if (Actor.isAtHome()) {
-                    throw error;
-                } else {
-                    this.log.warning(
-                        `Failed to fetch user data based on token, disabling proxy.`,
-                        { error },
-                    );
-                    return;
-                }
+        if (!token) return;
+        try {
+            const user = await Actor.apifyClient.user().get();
+            this.password = user.proxy?.password;
+        } catch (error) {
+            if (Actor.isAtHome()) {
+                throw error;
+            } else {
+                this.log.warning(`Failed to fetch user data using token`, {
+                    error,
+                });
             }
-
-            const { password } = proxy!;
-
-            this.password = password;
         }
     }
 
--- a/packages/apify/src/proxy_configuration.ts
+++ b/packages/apify/src/proxy_configuration.ts
@@ -418,52 +418,47 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
      * Checks if Apify Token is provided in env and gets the password via API and sets it to env
      */
     protected async _setPasswordIfToken(): Promise<void> {
-        const token = this.config.get('token');
+        if (!this.password) {
+            const token = this.config.get('token');
 
-        if (token) {
-            let proxy: UserProxy;
+            if (token) {
+                let proxy: UserProxy;
 
-            try {
-                const user = await Actor.apifyClient.user().get();
-                proxy = user.proxy!;
-            } catch (error) {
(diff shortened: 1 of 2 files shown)


packages/apify/src/proxy_configuration.ts
```diff
@@ -418,52 +418,47 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
      * Checks if Apify Token is provided in env and gets the password via API and sets it to env
      */
     protected async _setPasswordIfToken(): Promise<void> {
-        const token = this.config.get('token');
+        if (!this.password) {
```
B4nan, line 421: early return is better than this, now the whole method implementation is inside the if block.

```suggestion
        if (this.password) {
            return;
        }
```

packages/apify/src/proxy_configuration.ts
```diff
@@ -418,52 +418,47 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
...
+            if (!this.password) {
+                if (Actor.isAtHome()) {
+                    throw new Error(
+                        `Apify Proxy password must be provided using options.password or the "${APIFY_ENV_VARS.PROXY_PASSWORD}" environment variable. ` +
+                            `If you add the "${APIFY_ENV_VARS.TOKEN}" environment variable, ` +
+                            `while not providing "${APIFY_ENV_VARS.PROXY_PASSWORD}", the password will be automatically inferred.`,
+                    );
+                } else {
                     this.log.warning(
-                        'The Apify Proxy password you provided belongs to' +
```
B4nan, line 446: do we really want to remove this? you said this

> since only one is going to be taken into consideration.

but this was always the case, the warning was there to let user know there is something to pay attention to, we always "considered only one".

B4nan (changes requested):

stepskop:
> this is trying to solve https://apify.slack.com/archives/C0L33UM7Z/p1746448939282809 right? lets mention this in the PR description

Actually, I wasn't aware of this specific bug report :) . But yea this will probably fix the issue. I will mention it.

packages/apify/src/proxy_configuration.ts
stepskop: When running on Apify, the passwords will never match, since a special token is generated every run.  That leads to being logged for every run on the platform, and the current solution is exclusion in Mezmo, which we want to get rid of with these changes.

However, we can keep fetching and notifying the developers if `Actor.isAtHome()`. Is that what you mean?

packages/apify/src/proxy_configuration.ts
stepskop: Changed.

test/apify/proxy_configuration.test.ts
```diff
@@ -619,7 +619,7 @@ describe('Actor.createProxyConfiguration()', () => {
         getUserSpy.mockRestore();
     });
 
-    test('should show warning log', async () => {
+    test.skip('should show warning log', async () => {
```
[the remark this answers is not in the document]
stepskop: Yeah sure, I was planning to remove it, but you were quicker with the review ⚡.

B4nan (approved):

tobice (changes requested):
I think the overall logic can be simplified a bit, to better convey what it is actually doing. See the suggestion.

Also, the PR title is long and confusing :sweat_smile: 

Maybe: `fix: fetch proxy password from API only as fallback`?

packages/apify/src/proxy_configuration.ts
```diff
@@ -418,6 +418,7 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
      * Checks if Apify Token is provided in env and gets the password via API and sets it to env
      */
     protected async _setPasswordIfToken(): Promise<void> {
```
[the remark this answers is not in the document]
B4nan: For some reason, the method is protected, so technically, renaming it is a BC. 

(I guess it should be safe, but still.)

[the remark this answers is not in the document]
tobice: Ah, that's true... well, it's kind of your call here :smile: 

We could also keep it around, mark it as deprecated, and from within do `this.password = fetchPasswordFromApi()`.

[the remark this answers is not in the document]
B4nan: I would rather stay on the safe side and keep it around, marked as deprecated or have a TODO on it so we can rename and make it private in the next major (which we will hopefully get to later this year or early next year).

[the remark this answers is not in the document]
stepskop: We can keep the name and make the checks and warnings as @tobice proposed:

```typescript
    async initialize(): Promise<boolean> {
        if (this.usesApifyProxy) {
            if (!this.password) {
                 await this._setPasswordIfToken();
            }
            
            if (!this.password) {
                 // Do the logging here.
            }
            
            return this._checkAccess();
        }
        return true;
```

That makes the most sense to me.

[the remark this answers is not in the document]
tobice: Looks good 👍

packages/apify/src/proxy_configuration.ts
```diff
@@ -260,7 +260,27 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
...
+            if (!this.password) {
+                await this._setPasswordIfToken();
+            }
+
+            if (!this.password) {
+                if (Actor.isAtHome()) {
+                    throw new Error(
+                        `Apify Proxy password must be provided using options.password or the "${APIFY_ENV_VARS.PROXY_PASSWORD}" environment variable. ` +
+                            `If you add the "${APIFY_ENV_VARS.TOKEN}" environment variable, ` +
+                            `while not providing "${APIFY_ENV_VARS.PROXY_PASSWORD}", the password will be automatically inferred.`,
```
tobice, line 272: (nit) _Inferred_ sounds like a mathematical operation. Why not just fetched? :smile: 

Or even an alternative formulation:

```
`You can also provide your Apify token via the "${APIFY_ENV_VARS.TOKEN}" environment variable, so that the SDK can fetch the proxy password from Apify API.`
```

Dtto below.

packages/apify/src/proxy_configuration.ts
```diff
@@ -415,8 +435,9 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
     }
 
     /**
-     * Checks if Apify Token is provided in env and gets the password via API and sets it to env
+     * Checks if proxy password is provided in env, if not, fetches it from API using Apify Token
```
tobice, line 438: It does not check anything anymore...

```suggestion
     * Fetch & set the proxy password from Apify API if an Apify token is provided.
```

packages/apify/src/proxy_configuration.ts
```diff
@@ -440,31 +461,7 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
...
-                        `If you add the "${APIFY_ENV_VARS.TOKEN}" environment variable, the password will be automatically inferred.`,
-                );
-            } else {
-                this.log.warning(
-                    `No proxy password or token detected, running without proxy. To use Apify Proxy locally, ` +
-                        `provide options.password or "${APIFY_ENV_VARS.PROXY_PASSWORD}" environment variable. ` +
-                        `If you add the "${APIFY_ENV_VARS.TOKEN}" environment variable, the password will be automatically inferred.`,
-                );
-            }
+            this.password = password;
```
tobice, line 464: (nit) IMHO this whole function can be now simplified.

```
    protected async _setPasswordIfToken(): Promise<void> {
        const token = this.config.get('token');
        
        if (!token) {
            return;
        }
        
        try {
            const user = await Actor.apifyClient.user().get();
            this.password = user.proxy?.password;
        } catch (error) {
            if (Actor.isAtHome()) {
                throw error;
            } else {
                this.log.warning(`Failed to fetch user data using token`,  { error });
            }
        }
    }
```

Also `disabling proxy` is misleading within the scope of this function.

tobice (approved):
Approving, but let's fix those few outdated comments... and also consider using the opportunity and make `_setPasswordIfToken` even simpler.

packages/apify/src/proxy_configuration.ts
stepskop: Done.

packages/apify/src/proxy_configuration.ts
stepskop: Yeah, my bad. Fixed.

packages/apify/src/proxy_configuration.ts
stepskop: Used this with a slight addition.

PR #404, opened 2025-06-25
feat: add { isAborting: true } to 'persistState' event if run is aborted

metalwarrior665:
Feel free to improve the wording in the docs. Also I'm not sure if I don't need to change anywhere else.


barjin (approved):
Looks good to me, thanks! :+1: 

Worst case, this could cause double persist (in case the user is now listening on both 'aborting' and 'persistState' events to persist). The persist callback should be (by definition) idempotent, though, so this is IMO good to go.

metalwarrior665:
So in the end [this was solved on platform's side](https://apify.slack.com/archives/CD0SF6KD4/p1750932230121329) by adding `{ isAborting: true, isMigrating: false }` to aborting event

PR #409, opened 2025-07-06
feat: respect input schema defaults in `Actor.getInput()`

### LINKED ISSUE #287: Support input schema defaults locally in `Actor.getInput()`
Currently, the input schema is ignored locally, and to respect the defaults inside it, you need to use the Apify CLI to run the actor, which creates the INPUT.json file based on those defaults when missing.

Implement input schema support on the SDK level, so this works on the fly if the INPUT.json file is missing. If it's there, we should still fill in the missing fields that have defaults in the input schema.

The goal of this PR is to unify the local vs platform behavior. As a nice side effect, there shouldn't be any need to use the Apify CLI to run the actor, as `npm start` should behave the same with this implemented.

vladfrangu:
Way too overdue, closes #287

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -49,7 +49,7 @@ import { ChargingManager } from './charging.js';
 import { Configuration } from './configuration.js';
 import {
     getDefaultsFromInputSchema,
-    kNoActorInputSchemaDefined,
+    noActorInputSchemaDefinedMarker,
     readInputSchema,
 } from './input-schemas.js';
 import { KeyValueStore } from './key_value_store.js';
@@ -2328,7 +2328,7 @@ export class Actor<Data extends Dictionary = Dictionary> {
         const inputSchema = readInputSchema();
 
         // Don't emit warning if there is no input schema defined
-        if (inputSchema === kNoActorInputSchemaDefined) {
+        if (inputSchema === noActorInputSchemaDefinedMarker) {
             return input;
         }
 
--- a/packages/apify/src/input-schemas.ts
+++ b/packages/apify/src/input-schemas.ts
@@ -27,14 +27,17 @@ const readJSONIfExists = (path: string): Dictionary | null => {
     return null;
 };
 
-export const kNoActorInputSchemaDefined = Symbol.for(
+/**
+ * @ignore
+ */
+export const noActorInputSchemaDefinedMarker = Symbol.for(
     'apify.noActorInputSchemaDefined',
 );
 
 export const readInputSchema = ():
     | Dictionary
     | null
-    | typeof kNoActorInputSchemaDefined => {
+    | typeof noActorInputSchemaDefinedMarker => {
     const localConfig = readJSONIfExists(
         join(process.cwd(), ACTOR_SPECIFICATION_FOLDER, LOCAL_CONFIG_NAME),
     );
@@ -68,7 +71,7 @@ export const readInputSchema = ():
 
     // If we are in an Actor context, BUT we do not have an input schema defined, we want to skip the warning
     if (!localConfig?.input) {
-        return kNoActorInputSchemaDefined;
+        return noActorInputSchemaDefinedMarker;
     }
 
     return null;


packages/apify/src/input-schemas.ts
```diff
@@ -0,0 +1,70 @@
...
+function readJSONIfExists(path: string): Dictionary | null {
+    if (existsSync(path)) {
+        const content = readFileSync(path, 'utf8');
+        return JSON.parse(content);
+    }
+
+    return null;
+}
+
+export const readInputSchema = (): Dictionary | null => {
```
barjin: nit: can we make all of them (`readJSONIfExists`, `readInputSchema`, `getDefaultsFromInputSchema`) `function` (or make all of them `const` arrow function)?

packages/apify/src/input-schemas.ts
```diff
@@ -0,0 +1,70 @@
...
+    const localConfig = readJSONIfExists(
+        join(process.cwd(), ACTOR_SPECIFICATION_FOLDER, LOCAL_CONFIG_NAME),
```
barjin, line 42: `process.cwd()` will likely break if the Dockerfile entry point is not `npm (--prefix=app/) start`, but e.g. `node app/main.js`

But I'm not sure how else we would refer to these files :thinking: I guess it's okay to live with this, especially since on Platform, we load this via API.

barjin (approved):
Few nits, but looks alright to me otherwise, thank you!

packages/apify/src/actor.ts
```diff
@@ -2273,4 +2286,33 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+        const env = this.getEnv();
+        let inputSchema: Dictionary | undefined | null;
+
+        // On platform, we can get the input schema from the build data
+        if (this.isAtHome() && env.actorBuildId) {
+            const buildData = await this.apifyClient
+                .build(env.actorBuildId)
+                .get();
+
+            inputSchema = buildData?.actorDefinition?.input;
```
janbuchar: I would at least log an error if this is not present in the response.

packages/apify/src/actor.ts
```diff
@@ -2273,4 +2286,33 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+        } else {
+            // On local, we can get the input schema from the local config
+            inputSchema = readInputSchema();
```
janbuchar: It'd also be nice to log something if no input schema is found.

packages/apify/src/input-schemas.ts
janbuchar: I only gave it a quick look, but it looks like we don't copy the `.actor` directory in our Dockerfiles anyway, so for the "build a Docker image and run it outside of the platform" use case, the answer to "But I'm not sure how else we would refer to these files" is probably "tough luck lol" 🤷 

But if we add a nice error log when there's not input schema to be found, yeah, we can live with that.

packages/apify/src/actor.ts
```diff
@@ -2291,4 +2304,33 @@ export class Actor<Data extends Dictionary = Dictionary> {
             ].join('\n'),
         );
     }
+
+    private async insertDefaultsFromInputSchema<T extends Dictionary>(
```
B4nan: `infer` seems like a better word here

```suggestion
    private async inferDefaultsFromInputSchema<T extends Dictionary>(
```

packages/apify/src/input-schemas.ts
```diff
@@ -0,0 +1,70 @@
+// TODO: v0, move all this logic from here and apify-cli to input_schema module
```
B4nan: I'd rather see an issue instead of a TODO in code, TODOs always rot in there.

packages/apify/src/actor.ts
```diff
@@ -2291,4 +2304,33 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+        input: T,
+    ): Promise<T> {
+        // TODO: v0, move all this logic from here and apify-cli to input_schema module
```
B4nan: same here

packages/apify/src/actor.ts
```diff
@@ -2291,4 +2304,33 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+
+        const env = this.getEnv();
+        let inputSchema: Dictionary | undefined | null;
+
+        // On platform, we can get the input schema from the build data
+        if (this.isAtHome() && env.actorBuildId) {
```
B4nan: why do we even need this? this change was only supposed to be dealing with the local development, on platform you always get the whole input from the API

packages/apify/src/input-schemas.ts
```diff
@@ -0,0 +1,70 @@
...
+import process from 'node:process';
+
+import type { Dictionary } from '@crawlee/utils';
+
+const DEFAULT_INPUT_SCHEMA_PATHS = [
+    ['.actor', 'INPUT_SCHEMA.json'],
+    ['INPUT_SCHEMA.json'],
+    ['.actor', 'input_schema.json'],
+    ['input_schema.json'],
+];
```
B4nan, line 15: the name of the input schema file is configurable via `.actor.json`, ideally we'd get it from there

https://github.com/apify/actor-templates/blob/master/templates/js-crawlee-puppeteer-chrome/.actor/actor.json#L10

packages/apify/src/actor.ts
vladfrangu: Oh right, platform won't need it since it deals with it right?

B4nan: Better try it out to be sure, but afaik the merging is done on the platform side, maybe when the input is being saved. The point of this PR should be to have the same behavior locally, without the need to use `apify run`.

packages/apify/src/input-schemas.ts
vladfrangu: I mean outside process.cwd, we have no way of finding these files... Maybe if we just walk up from process.cwd but I would rather... not :D

vladfrangu:
If this is all good from you guys, lets merge it and release it!

packages/apify/src/input-schemas.ts
```diff
@@ -0,0 +1,89 @@
...
+const readJSONIfExists = (path: string): Dictionary | null => {
+    if (existsSync(path)) {
+        const content = readFileSync(path, 'utf8');
+        return JSON.parse(content);
+    }
+
+    return null;
+};
+
+export const kNoActorInputSchemaDefined = Symbol.for(
```
B4nan: this is really weird naming. also, do we really need to export this symbol? if so, it should be marked as `@ignore` so it wont end up in the docs

vladfrangu: I just used naming similar to what I've seen in node (and we need it exported to check for it in actor.ts). I'll slap on the ignore tag in a sec

B4nan: i think we already discussed this some time ago. its weird, confusing, and unnecessary, lets not adopt it just because people maintaining node use it.

btw it looks like this file is not exported from the barrel so its fine i guess

B4nan (approved):

PR #413, opened 2025-07-11
fix: tagline overlap

katzino:
fixes tagline overlap (reported [here](https://apify.slack.com/archives/C0L33UM7Z/p1752206692759679))


B4nan:
same here, please keep in mind that such commits should never be `feat/fix`, the only exception is the apify-docs repo, which doesnt represent any library. please pay attention to this going forward (and spread the word in your team).

PR #414, opened 2025-07-14
docs: enable experimental faster features in Docusaurus config

barjin:
Switching to SWC from `babel` seems to fix the crashing API documentation issue

related to https://github.com/apify/apify-docs/issues/1686


B4nan:
@barjin I wonder why it works fine here with NPM, but in apify-docs you had to switch to yarn to be able to bump docusaurus to v3.8 (or was it some other issue?). There must be a way to make this work without switching the package manager in there.

PR #415, opened 2025-07-21
chore: add `userIsPaying` to the env var interface

stepskop:
Adds `APIFY_USER_IS_PAYING` environmental variable as `userIsPaying` to the `ApifyEnv` interface

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -112,6 +112,11 @@ export interface ApifyEnv {
     userId: string | null;
 
     /**
+     * If set to "1", the user calling the Actor is as paying user (APIFY_USER_IS_PAYING)
+     */
+    userIsPaying: string | null;
+
+    /**
      * Authentication token representing privileges given to the Actor run,
      * it can be passed to various Apify APIs (APIFY_TOKEN)
      */


packages/apify/src/actor.ts
```diff
@@ -111,6 +111,11 @@ export interface ApifyEnv {
      */
     userId: string | null;
 
+    /**
+     * If set to "1", the user calling the Actor is paying user (APIFY_USER_IS_PAYING)
+     */
+    userIsPaying: string | null;
```
barjin, line 117: Let's follow the other binary envvar name (`isAtHome`) and go for `isUserPaying`?

`is` prefix IMO communicates the binary value nicely.

packages/apify/src/actor.ts
```diff
@@ -111,6 +111,11 @@ export interface ApifyEnv {
      */
     userId: string | null;
 
+    /**
+     * If set to "1", the user calling the Actor is paying user (APIFY_USER_IS_PAYING)
```
barjin: https://github.com/apify/apify-docs/pull/1706

Let's use the same wording across all docs

packages/apify/src/actor.ts
barjin: Ah, I see, this is just a camel-cased name of the envvar. too late for changes then :)

packages/apify/src/actor.ts
stepskop: 👍 Fixed

packages/apify/src/actor.ts
stepskop: True, on the other hand, it follows the same pattern:

`APIFY_IS_AT_HOME` -> `isAtHome`
`APIFY_USER_IS_PAYING` -> `userIsPaying`

vladfrangu (approved):
Leaving final review to Jindra

barjin (approved):
lgtm, thanks! :rocket:

PR #420, opened 2025-07-30
fix: force quit stuck `Actor.exit()` calls

barjin:
Encountered in https://console.apify.com/view/runs/ZlPUJOjFTsIHzH8dP

image

`WARN  Main crawler finished successfully, exiting the run...` is the last log line before the `Actor.exit()` call. The internals of the call seem to be blocking execution for ~1 hour, before the Platform migrates.

These changes place the `process.exit()` call into the event loop before any potentially time-consuming operations.

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -329,7 +329,7 @@ export const EXIT_CODES = {
  */
 export class Actor<Data extends Dictionary = Dictionary> {
     /** @internal */
-     
+    // eslint-disable-next-line no-use-before-define -- self-reference
     static _instance: Actor;
 
     /**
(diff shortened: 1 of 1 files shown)


packages/apify/src/actor.ts
```diff
@@ -553,6 +553,15 @@ export class Actor<Data extends Dictionary = Dictionary> {
         log.debug(
             `Waiting for all event listeners to complete their execution (with ${options.timeoutSecs} seconds timeout)`,
         );
+
+        if (options.exit) {
+            // `addTimeoutToPromise` is a cooperative timeout. This ensures that the process exits
```
janbuchar, line 558: Well, if something hogs the event loop with something CPU-intensive, the setTimeout won't fire, but this surely is better than the previous state

PR #429, opened 2025-08-18
docs: Fix nodejs versions

patrikbraborec:


docs/guides/docker_images.mdx
```diff
@@ -39,18 +39,18 @@ Each image is tagged with up to 2 version tags, depending on the type of the ima
 
 ### Node.js versioning
 
-Our images are built with multiple Node.js versions to ensure backwards compatibility. Currently, Node.js **versions 16 and 18 are supported** (legacy versions still exist, see DockerHub). To select the preferred version, use the appropriate number as the image tag.
+Our images are built with multiple Node.js versions to ensure backwards compatibility. Currently, Node.js **version 24 is supported** (legacy versions still exist, see DockerHub). To select the preferred version, use the appropriate number as the image tag.
```
B4nan: afaik we still do support v18, v20 and v22

we will likely drop v18 soon

cc @vladfrangu

vladfrangu: I think I already dropped 18 because its EoL but I don't remember if we added 24 yet. That said docs should probably not mention a specific list of versions but just "currently actively maintained versions of node" or automate it in someway

patrikbraborec: @vladfrangu we have it in the rest of the document so I would keep it listed.

Let's then mention v20, v22 and v24?

vladfrangu: Pain. Sure! But in the future I feel like we should not mention versions imo but more wavey-hand-supported-versions (but others might disagree in which case I defer to them)

B4nan (approved):
we have the same issue here:

https://crawlee.dev/js/docs/guides/docker-images#nodejs-versioning

PR #436, opened 2025-08-25
docs: Shorten long generated descriptions

HonzaTuron:
Part of apify/apify-docs#1783

As some descriptions contains code and are multiline (even 20 or more lines), I've used description prop for affected files.

### DIFF
--- a/website/versioned_docs/version-2.3/api/Configuration.md
+++ b/website/versioned_docs/version-2.3/api/Configuration.md
@@ -1,6 +1,7 @@
 ---
 id: configuration
 title: Configuration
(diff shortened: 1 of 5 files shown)


barjin (approved):
Note that you're editing docs for the ~4 y/o version of Apify SDK. The "current" docs are rendered from the JSDoc comments in the code (e.g. see https://docs.apify.com/sdk/js/reference/class/Log). For >3.0 versions, these are versioned in the `api-typedoc.json` files.

Approving, since the changes look harmless :+1:  but they might not be enough.

PR #451, opened 2025-09-01
fix: consistent parameters for platform event listeners

### LINKED ISSUE #405: Actor.reboot calls listeners without parameter
### Issue description

Actor.reboot calls all listeners attached to `persistState` and `migrating`. [Our code examples expect data to be an object](https://docs.apify.com/platform/actors/development/programming-interface/system-events#handle-system-events) but currently reboot calls listeners without parameters so the example would crash.

At minimum we should send empty object but ideally we figure out some values that are consistent. Currently, there can be `{ isMigrating: boolean }` (and I recently proposed also `{ isAborting: boolean }` so I guess we should re-send these. We could include something like `calledByReboot: boolean` so users can not persist again after the same listener is called again but I wouldn't do that yet, we might want to rethink how this works (will make separate issue)

### Code sample

```TypeScript
Actor.on('persistState', async (data) => {
    // This listener is called again by reboot but 'data' is undefined
    if (data.isMigrating) {
        await Actor.reboot()
    }
})
```

### Package version

latest

### Node.js version

latest

### Operating system

_No response_

### Actor or run link

_No response_

### I have tested this on the `next` release

_No response_

### Other context

_No response_

barjin:
Closes #405

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -834,12 +834,12 @@ export class Actor<Data extends Dictionary = Dictionary> {
             ...this.config
                 .getEventManager()
                 .listeners(EventType.PERSIST_STATE)
-                .map(async (x) => x()),
+                .map(async (x) => (x as any)({})),
             // `migrating` to pause Apify crawlers
             ...this.config
                 .getEventManager()
                 .listeners(EventType.MIGRATING)
-                .map(async (x) => x()),
+                .map(async (x) => (x as any)({})),
         ]);
 
         const runId = this.config.get('actorRunId')!;


barjin:
```javascript
Actor.on('persistState', async (data) => {
    if (data.isMigrating) {
        await Actor.reboot()
    }
});
```

Before this change, `Actor.reboot()` would call the `persistState` listener with no parameter, causing `data.isMigrating` to fail on accessing a property of `undefined`.

packages/apify/src/actor.ts
```diff
@@ -834,12 +834,12 @@ export class Actor<Data extends Dictionary = Dictionary> {
             ...this.config
                 .getEventManager()
                 .listeners(EventType.PERSIST_STATE)
-                .map(async (x) => x()),
+                .map(async (x) => (x as any)({})),
```
janbuchar, line 837: This is horrible. But I guess it's the right approach to fix an actual issue first and worry about making the event manager reasonably typesafe later.

barjin:
Related to https://github.com/apify/crawlee/pull/3027/files

PR #456, opened 2025-09-04
chore(ci): fix failing e2e tests

### LINKED ISSUE #363: Fix failing e2e tests
https://github.com/apify/apify-sdk-js/actions/runs/13232329263/job/36931357642

- I tried running the failing tests locally, which did not reproduce the issues
- maybe there is some problem with Puppeteer versions?
- it has been failing more or less consistently since the beginning of January

barjin:
Fixes failing E2E tests (updates dependencies and edits the test assertions based on the current contents of `crawlee.dev`).

Closes #363

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
                 "@apify/consts": "^2.29.0",
                 "@apify/eslint-config": "^1.0.0",
                 "@apify/input_secrets": "^1.2.0",
+                "@apify/log": "^2.5.22",
                 "@apify/tsconfig": "^0.1.0",
                 "@commitlint/config-conventional": "^19.2.2",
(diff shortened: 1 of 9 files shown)


test/e2e/runSdkTests.mjs
```diff
@@ -51,6 +51,8 @@ async function runWorker(dirName) {
                 log.info('Captured stderr:');
                 process.stderr.write(err);
             }
+
+            process.exit(exitCode);
```
barjin, line 55: If we don't exit the main process on failing test, CI job will succeed with :heavy_check_mark: even if the tests are failing.

See https://github.com/apify/apify-sdk-js/actions/runs/17490187761 where the `Actor.charge()` test is failing (for what seems to be a good reason), but the job is green.

PR #471, opened 2025-09-18
fix: ensure the `Actor` instance is initialized while calling its methods

barjin:
Adds a warning on a non-initialized Actor on `useState`, `exit`, `reboot` and `setStatusMessage` method calls.


barjin:
Side note - `Actor.exit()`, `Actor.reboot()` and `Actor.setStatusMessage()` all use some components initialized in `Actor.init()` (storage client and event manager). Should we perhaps add the warnings to those too?

B4nan:
> Side note - Actor.exit(), Actor.reboot() and Actor.setStatusMessage()

Good point, `exit` without `init` makes no sense, same for `setStatusMessage`, and likely too for the `reboot` method.

B4nan (approved):

PR #479, opened 2025-09-26
fix: Correct link to documentation in `getEnv()` method

fnesveda:
The original link didn't actually contain the list of environment variables, this new one does.


fnesveda:
Thanks, good point 👍

PR #483, opened 2025-10-07
perf: Use Apify-provided environment variables to obtain PPE pricing information

### LINKED ISSUE #481: Load Actor pricing and charged events from env vars
### Feature

There are now two new env vars: `APIFY_CHARGED_ACTOR_EVENT_COUNTS` and `APIFY_ACTOR_PRICING_INFO` plus there is still the existing `ACTOR_MAX_TOTAL_CHARGE_USD`. These can be used to replace the pricing data currently fetched from GET Run endpoint. The types are the same as in the API just with `description` stripped out.

The env vars are updated after the resurrect, so the behavior is consistent with refetching the GET Run endpoint now.

More details in [Slack thread](https://apify.slack.com/archives/C010Q0FBYG3/p1759217483100159?thread_ts=1758088160.849999&cid=C010Q0FBYG3)

### Motivation

1. Lower start latency by skipping the HTTP call
2. More resistant to API rate limits. For some reason the charging manager init was vulnerable to crash the Actor if the per-account API rate limit was reached.
3. This can make it easier to later implement local testing. The `ACTOR_MAX_TOTAL_CHARGE_USD` can be used locally and we could add a CLI helper to inject pricing env vars from local JSON files.

Don't think there is any complication unless the platform somehow messes up the stringification :) 

_No response_

### Other context

_No response_

janbuchar:
- closes #481

### DIFF
--- a/packages/apify/src/charging.ts
+++ b/packages/apify/src/charging.ts
@@ -97,20 +97,6 @@ export class ChargingManager {
         this.purgeChargingLogDataset = configuration.get('purgeOnStart');
         this.useChargingLogDataset = configuration.get('useChargingLogDataset');
 
-        if (this.useChargingLogDataset && this.isAtHome) {
-            throw new Error(
-                'Using the ACTOR_USE_CHARGING_LOG_DATASET environment variable is only supported in a local development environment',
-            );
-        }
-
-        if (configuration.get('testPayPerEvent')) {
-            if (this.isAtHome) {
-                throw new Error(
-                    'Using the ACTOR_TEST_PAY_PER_EVENT environment variable is only supported in a local development environment',
-                );
-            }
-        }
-
         this.apifyClient = apifyClient;
     }
 
@@ -170,6 +156,21 @@ export class ChargingManager {
      * Initialize the ChargingManager by loading pricing information and charging state via Apify API.
      */
     async init(): Promise<void> {
+        // Validate config - it may have changed since the instantiation
+        if (this.useChargingLogDataset && this.isAtHome) {
+            throw new Error(
+                'Using the ACTOR_USE_CHARGING_LOG_DATASET environment variable is only supported in a local development environment',
+            );
+        }
+
+        if (this.configuration.get('testPayPerEvent')) {
+            if (this.isAtHome) {
+                throw new Error(
+                    'Using the ACTOR_TEST_PAY_PER_EVENT environment variable is only supported in a local development environment',
+                );
+            }
+        }
+
         // Retrieve pricing information
         const { pricingInfo, chargedEventCounts, maxTotalChargeUsd } =
             await this.fetchPricingInfo();


packages/apify/src/charging.ts
```diff
@@ -117,12 +135,48 @@ export class ChargingManager {
...
+    }
+
     /**
      * Initialize the ChargingManager by loading pricing information and charging state via Apify API.
      */
     async init(): Promise<void> {
-        this.chargingState = {};
-
         // Retrieve pricing information
         if (this.isAtHome) {
```
metalwarrior665: We should be able to get rid of this with a future refactor so we could inject the pricing locally

metalwarrior665 (approved):
LGTM

packages/apify/src/charging.ts
janbuchar: I'm not sure what you're talking about, you can already change the pricing locally using the environment variable - do you have something else in mind?

metalwarrior665: My bad, just overlooked what is running inside the ` if (this.isAtHome) {`

packages/apify/src/charging.ts
```diff
@@ -117,12 +135,47 @@ export class ChargingManager {
...
+    ) {
+        this.chargingState = {};
+
+        for (const [eventName, chargeCount] of Object.entries(
+            chargedEventCounts ?? {},
+        )) {
+            this.chargingState[eventName] = {
+                chargeCount,
+                totalChargedAmount:
+                    chargeCount * (this.pricingInfo[eventName]?.price ?? 0),
```
barjin, line 206: nit: you have to call the new `load*` methods in this specific order, as `loadChargedEventCounts` reads variables initialized by `loadPricingInfo` (temporal coupling)

packages/apify/src/charging.ts
```diff
@@ -131,40 +184,25 @@ export class ChargingManager {
...
-            if (run.pricingInfo?.pricingModel === 'PAY_PER_EVENT') {
-                for (const [eventName, eventPricing] of Object.entries(
-                    run.pricingInfo.pricingPerEvent.actorChargeEvents,
-                )) {
-                    this.pricingInfo[eventName] = {
-                        price: eventPricing.eventPriceUsd,
-                        title: eventPricing.eventTitle,
-                    };
+            if (
+                this.chargingState === undefined ||
```
barjin: how about removing the `load*` method calls from the constructor, running them only once in the `init` method (at one call site) and conditionally loading the config with something like

```javascript
async function getPricingInfo() {
   if (
        configuration.get('actorPricingInfo') &&
        configuration.get('chargedEventCounts')
    ) return { pricingInfo: JSON.parse(), ...

    const run = await this.apifyClient.run(this.actorRunId).get();
    ...
}
```

This would allow us to load the pricing info + events at one point in `init` like

```
...
const { pricingInfo ... } = await getPricingInfo();
this.loadPricingInfo( pricingInfo );
...
```

which is IMO cleaner than the current approach, which is conditional based on the current object state

packages/apify/src/charging.ts
```diff
@@ -136,33 +151,59 @@ export class ChargingManager {
...
+     * Initialize the ChargingManager by loading pricing information and charging state via Apify API.
+     */
+    async init(): Promise<void> {
+        // Retrieve pricing information
+        const { pricingInfo, chargedEventCounts, maxTotalChargeUsd } =
+            await this.fetchPricingInfo();
+
+        if (this.configuration.get('testPayPerEvent')) {
+            this.pricingModel = 'PAY_PER_EVENT';
+        } else {
```
barjin, line 180: doing e.g. this:
```typescript
Actor.config.set('testPayPerEvent', true);
await Actor.init();
await Actor.charge({
    eventName: 'my-event',
    count: 1,
});
await Actor.exit();
```

Will allow enabling the `testPayPerEvent` mode even on the platform. This wouldn't have been possible before due to [this check](https://github.com/apify/apify-sdk-js/pull/483/files#diff-a5f685af5ad54bd14b531b72d14468a6f62ba0b155bb37851c1a69979115b160R107). 

Is this change intentional? Can we move the `isAtHome` check here to `init()`? What was the original reason for the check anyway?

janbuchar: The reason for the check is that you shouldn't do weird stuff like running the PPE test mode on Apify. I haven't thought of people being able to modify `Actor.config` like this. Not sure if it's a problem but I'll patch it up.

barjin: Yeah, I agree it's a rather academic example. It's unlikely to be an exploitable issue, it just stood out to me.

PR #504, opened 2025-11-10
fix: Fix Actor.charge behavior when the budget is overdrawn

janbuchar:
- Porting over the fix and tests from https://github.com/apify/apify-sdk-python/pulls/668

### DIFF
--- a/packages/apify/src/charging.ts
+++ b/packages/apify/src/charging.ts
@@ -425,18 +425,14 @@ export class ChargingManager {
             return Infinity;
         }
 
+        // The raw number of events allowed by the budget
+        const unroundedResult =
+            (this.maxTotalChargeUsd - this.calculateTotalChargedAmount()) /
+            price;
+
         // First round as Math.floor(4.9999999999999999) will incorrectly return 5
-        return Math.max(
-            0,
-            Math.floor(
-                Number(
-                    (
-                        (this.maxTotalChargeUsd -
-                            this.calculateTotalChargedAmount()) /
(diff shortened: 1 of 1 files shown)


packages/apify/src/charging.ts
```diff
@@ -426,13 +426,16 @@ export class ChargingManager {
...
+        return Math.max(
+            0,
+            Math.floor(
+                Number(
+                    (
+                        (this.maxTotalChargeUsd -
+                            this.calculateTotalChargedAmount()) /
+                        price
+                    ).toFixed(4),
+                ),
```
B4nan: lets introduce some local vars so this is less ugly

test/apify/charging.test.ts
```diff
@@ -1,3 +1,5 @@
+import { afterEach, beforeEach, describe, expect, test } from 'vitest';
```
B4nan, line 1: is this really needed? i thought we have globals enabled for vitest, other tests dont do it

B4nan (approved):

test/apify/charging.test.ts
janbuchar: It's not, I just prefer the explicit imports and I added them because of force of habit. Should I remove it?

B4nan: Up to you, I'd rather not use them selectively (and all the other tests dont use them). But looking at crawlee, we already have this inconsistency there 🙃 not a huge deal

PR #516, opened 2025-11-25
feat: add `ACTOR_PERMISSION_LEVEL` to env var interface

stepskop:
This adds the `ACTOR_PERMISSION_LEVEL` environmental variable as `actorPermissionLevel` to the `ApifyEnv` interface.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,9 +54,9 @@
             "license": "MIT"
         },
         "node_modules/@apify/consts": {
-            "version": "2.47.0",
-            "resolved": "https://registry.npmjs.org/@apify/consts/-/consts-2.47.0.tgz",
-            "integrity": "sha512-VFVrNn/S3bLPrS3v9x7Td5b8YtxPbrepuXLWu27n06T34qQeogysbnAWIBNnTQ59qRKlxrAeFR4ezI+fRXTMNQ==",
+            "version": "2.47.1",
+            "resolved": "https://registry.npmjs.org/@apify/consts/-/consts-2.47.1.tgz",
+            "integrity": "sha512-psuf1uZ8E/9YV6cQXWGX18D7k9yCJPOSCdWI3Qgy0cx6yjckgafMBEtIK+o4BlGid58sAwxBC2dtcH7iIYj5uw==",
             "license": "Apache-2.0"
         },
         "node_modules/@apify/datastructures": {


packages/apify/src/actor.ts
```diff
@@ -497,6 +502,10 @@ export class Actor<Data extends Dictionary = Dictionary> {
         // reset global config instance to respect APIFY_ prefixed env vars
         CoreConfiguration.globalConfig = Configuration.getGlobalConfig();
 
+        log.info('Permissions', {
+            actorPermissionLevel: this.getEnv().actorPermissionLevel,
+        });
```
stepskop: I am not sure if this should not be somehow conditional, as most of the time it would be defined only when run on Apify. Locally, it won't be defined if the user doesn't do so.

However, I can imagine some cases when my code depends on the permission level, so I am passing it "artificially" even locally to test things out.

tobice: I was actually thinking that we'd print it via worker. That would be more universal 🤔

stepskop: I thought that the original intention was to put it in the **system message** (= SDK), but when implementing it, I decided not to put it there, as the system message consists of more "low level" information.

I guess we can log it through Worker.

tobice: I thought **system message** means worker :sweat_smile:

barjin: +1 for logging this elsewhere, locally this IMO doesn't add much value. It's either unset, i.e., this line will print `Permissions { actorPermissionLevel: null }`, or I pass it manually = I did that intentionally, I know what permission level I'm on.

packages/apify/src/actor.ts
```diff
@@ -106,6 +106,11 @@ export interface ApifyEnv {
      */
     actorBuildId: string | null;
 
+    /**
+     * [Permission level](https://docs.apify.com/platform/actors/development/permissions) the Actor is run under. (ACTOR_PERMISSION_LEVEL)
+     */
+    actorPermissionLevel: string | null;
```
barjin: Since this is an enum, I'm wondering if we could type this strictly (`'LIMITED_PERMISSIONS' | 'FULL_PERMISSIONS' | null`). 

It would help users with adoption, but might require some well-placed casts.

packages/apify/src/actor.ts
danpoletaev: Would also vote for worker or in worst case maybe add it to `log.info('System info', getSystemInfo());`

stepskop: I wouldn't add it to the system message as I said here: https://github.com/apify/apify-sdk-js/pull/516#discussion_r2564542183

Lets move to to Worker 👍

packages/apify/src/actor.ts
stepskop: I typed it using the `ACTOR_PERMISSION_LEVEL` from `@apify/consts`

PR #518, opened 2025-11-28
feat: Add `inherit` timeout option for `Actor.call` and `Actor.start`

Pijukatel:
-  Add `RemainingTime` timeout option for `Actor.call`, `Actor.start` and `Actor.callTask`. This will set the timeout of another `Actor` run to whatever is left from the timeout of the current `Actor` run

Related to: https://github.com/apify/apify-client-js/issues/632
Python implementation: https://github.com/apify/apify-sdk-python/pull/473

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,8 +51,7 @@
             "version": "0.9.23",
             "resolved": "https://registry.npmjs.org/@acemir/cssom/-/cssom-0.9.23.tgz",
             "integrity": "sha512-2kJ1HxBKzPLbmhZpxBiTZggjtgCwKg1ma5RHShxvd6zgqhDEdEkzpiwe7jLkI2p2BrZvFCXIihdoMkl1H39VnA==",
-            "license": "MIT",
-            "peer": true
+            "license": "MIT"
         },
         "node_modules/@apify/consts": {
             "version": "2.47.1",
@@ -170,7 +169,6 @@
             "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-4.0.5.tgz",
             "integrity": "sha512-lMrXidNhPGsDjytDy11Vwlb6OIGrT3CmLg3VWNFyWkLWtijKl7xjvForlh8vuj0SHGjgl4qZEQzUmYTeQA2JFQ==",
             "license": "MIT",
-            "peer": true,
             "dependencies": {
                 "@csstools/css-calc": "^2.1.4",
                 "@csstools/css-color-parser": "^3.1.0",
@@ -184,7 +182,6 @@
             "resolved": "https://registry.npmjs.org/@asamuzakjp/dom-selector/-/dom-selector-6.7.4.tgz",
             "integrity": "sha512-buQDjkm+wDPXd6c13534URWZqbz0RP5PAhXZ+LIoa5LgwInT9HVJvGIJivg75vi8I13CxDGdTnz+aY5YUJlIAA==",
             "license": "MIT",
-            "peer": true,
             "dependencies": {
                 "@asamuzakjp/nwsapi": "^2.3.9",
                 "bidi-js": "^1.0.3",
@@ -197,8 +194,7 @@
             "version": "2.3.9",
             "resolved": "https://registry.npmjs.org/@asamuzakjp/nwsapi/-/nwsapi-2.3.9.tgz",
             "integrity": "sha512-n8GuYSrI9bF7FFZ/SjhwevlHc8xaVlb/7HmHelnc/PZXBD2ZR49NnN9sMMuDdEGPeeRQ5d0hqlSlEpgCX3Wl0Q==",
-            "license": "MIT",
-            "peer": true
+            "license": "MIT"
         },
         "node_modules/@babel/code-frame": {
             "version": "7.27.1",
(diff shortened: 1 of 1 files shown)


packages/apify/src/actor.ts
```diff
@@ -2319,6 +2333,23 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+     */
+    private _getRemainingTime(): number | undefined {
+        const env = this.getEnv();
+        if (env.isAtHome === '1' && env.timeoutAt !== null) {
+            return env.timeoutAt.getTime() - Date.now();
+        }
+        log.warning(
+            'Returning `undefined` instead of remaining time. Using `RemainingTime` argument is only possible when ' +
+                'the Actor is running on the Apify platform and when the timeout for the Actor run is set. ' +
+                `${env.isAtHome}, ${env.timeoutAt}`,
```
barjin: Given the conditional check above, I suppose this will always print `undefined, null`, right? If so, it IMO doesn't bring any value, and I'd just drop it

```suggestion
```

packages/apify/src/actor.ts
```diff
@@ -221,11 +221,18 @@ export interface ApifyEnv {
...
      * User API token that is used to run the Actor. By default, it is taken from the `APIFY_TOKEN` environment variable.
      */
     token?: string;
+    /**
+     * Timeout for the Actor run in seconds, or `'RemainingTime'`.
+     *
+     * Using `RemainingTime` will set timeout of the other Actor run to the time
+     * remaining from this Actor run timeout.
+     */
+    timeout?: number | 'RemainingTime';
```
barjin: The Pascal-cased string constant doesn't feel right, format-wise. How about `'inherit'`?

Pijukatel: Not sure about the common casing of string literals in JS, but this is the same literal as in the Python implementation. So, unless there is some name suggestion that is better by a lot, then I would prefer to keep it consistent.

B4nan: I also like `inherit` much better.

janbuchar: So what do we do with the Python version now? Deprecate `RemainingTime` in favor of `inherit`?

packages/apify/src/actor.ts
```diff
@@ -227,11 +221,18 @@ export interface ApifyEnv {
...
+export interface CallOptions extends Omit<ActorCallOptions, 'timeout'> {
     /**
      * User API token that is used to run the Actor. By default, it is taken from the `APIFY_TOKEN` environment variable.
      */
     token?: string;
+    /**
+     * Timeout for the Actor run in seconds, or `'inherit'`.
+     *
+     * Using `inherit` will set timeout of the other Actor run to the time
+     * remaining from this Actor run timeout.
```
janbuchar:
```suggestion
     * Using `inherit` will set timeout of the newly started Actor run to the time
     * remaining until this Actor run times out so that the new run does not outlive this one.
```

packages/apify/src/actor.ts
```diff
@@ -2325,6 +2333,22 @@ export class Actor<Data extends Dictionary = Dictionary> {
         );
     }
 
+    /**
+     * Get time remaining from the Actor run timeout. Returns `undefined` if not on an Apify platform or the current
+     * run was started without a timeout.
+     */
+    private _getRemainingTime(): number | undefined {
+        const env = this.getEnv();
+        if (env.isAtHome === '1' && env.timeoutAt !== null) {
```
janbuchar:
```suggestion
        if (Boolean(env.isAtHome) && env.timeoutAt !== null) {
```

packages/apify/src/actor.ts
```diff
@@ -2325,6 +2333,22 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+            return env.timeoutAt.getTime() - Date.now();
+        }
+        log.warning(
+            'Returning `undefined` instead of remaining time. Using `inherit` argument is only possible when ' +
```
janbuchar: The user reading this message won't understand what "Returning `undefined`" means. I think we could just remove the first sentence and keep the second one.

packages/apify/src/actor.ts
B4nan: should be just `this.isAtHome()`

packages/apify/src/actor.ts
B4nan: I'd be up for that, what do you think about the naming here?

packages/apify/src/actor.ts
janbuchar: Oh, we're in `Actor` :facepalm: you're right

packages/apify/src/actor.ts
janbuchar: Yeah, I don't mind it :slightly_smiling_face:

Pijukatel: https://github.com/apify/apify-sdk-python/issues/697

packages/apify/src/actor.ts
Pijukatel: Ok, removed.

packages/apify/src/actor.ts
Pijukatel: Using `this.isAtHome()`

PR #522, opened 2025-12-08
feat: Enable log redirection and update docs

Pijukatel:
- Bump the `apify-client` dependency to enable log redirection.
- Add docs.

### DIFF
--- a/packages/apify/src/actor.ts
+++ b/packages/apify/src/actor.ts
@@ -1833,7 +1833,7 @@ export class Actor<Data extends Dictionary = Dictionary> {
     static async start(
         actorId: string,
         input?: Dictionary,
-        options: CallOptions = {},
+        options: StartOptions = {},
     ): Promise<ClientActorRun> {
         return Actor.getDefaultInstance().start(actorId, input, options);
     }


docs/guides/log_redirection.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+In some situations, one Actor is going to start one or more other Actors and wait for them to finish and produce some results. In such cases, you might want to redirect the logs of the started Actors' runs back to the parent Actor run, so that you can see the progress of the started Actors' runs in the parent Actor's logs. This guide will show possibilities on how to do it.
+
+### Redirecting logs from `Actor.call`
+
+Typical use case for log redirection is to call another Actor using the [`Actor.call()`](/reference/class/Actor#call) method. This method has an optional logger argument, which is by default set to the `default` literal. This means that the logs of the called Actor will be automatically redirected to the parent Actor's logs with default formatting and filtering. If you set the logger argument to `null`, then no log redirection happens. The third option is to pass your own `Log` instance with the possibility to define your specific logging. Below you can see those three possible ways of log redirection when starting another Actor run through Actor.call.
+
+```javascript
+import { Actor } from 'apify';
+import {LoggerActorRedirect} from 'apify-client';
```
barjin:
```suggestion
import { LoggerActorRedirect } from 'apify-client';
```

please lint / format the examples (this might not be the only formatting issue)

docs/guides/log_redirection.mdx
```diff
@@ -0,0 +1,64 @@
...
+// No log redirection
+await Actor.call(actorId, input, {log:null});
+
+// Custom log redirection
+await Actor.call(actorId, input, {log: new Log({ level: LEVELS.DEBUG, prefix: 'customPrefix', logger: new LoggerActorRedirect()});
+
+await Actor.exit();
+```
+
+Each default redirect log entry will have a specific format. After the timestamp, it will contain cyan colored text that will contain the redirect information - the other Actor's name and the run ID. The rest of the log message will be printed in the same manner as the parent Actor's log is configured.
```
barjin: > After the timestamp

The timestamp is only added by the Platform, so locally, this message will start with the colored redirect information, right?

But even if so, I don't really see a reason to change this, this is IMO too minuscule to confuse anybody too much.

docs/guides/log_redirection.mdx
```diff
@@ -0,0 +1,64 @@
...
+### Redirecting logs from already running Actor run
+
+In some cases, you might want to connect to an already running Actor run and redirect its logs to your current Actor run. This can be done using the <ApiLink to="class/ApifyClient">`ApifyClient`</ApiLink> and getting the streamed log from a specific Actor run. You can then control the log redirection manually by explicitly calling `start` and `stop` methods.
+
+You can further decide whether you want to redirect just new logs of the ongoing Actor run, or if you also want to redirect historical logs from that Actor's run, so all logs it has produced since it was started. This can be controlled by the `fromStart` option of the `getStreamedLog()` method. If you set it to `true`, all historical logs will be redirected first, and then new logs will be redirected as they are produced. If you set it to `false`, only new logs will be redirected.
+
+```javascript
+import { Actor } from 'apify';
+
+const actorRunId = 'someRunId'; // ID of actor you want to call
```
barjin:
```suggestion
const actorRunId = 'someRunId'; // ID of the run you want to connect to
```

packages/apify/src/actor.ts
```diff
@@ -718,7 +716,7 @@ export class Actor<Data extends Dictionary = Dictionary> {
     async start(
         actorId: string,
         input?: unknown,
-        options: CallOptions = {},
+        options: StartOptions = {},
```
barjin, line 719: This is curious, nice catch 👀 

I see that `StartOptions` is not larger than `CallOptions`, so this change might cause some type errors though. If the user was e.g. using the [`waitSecs`](https://github.com/apify/apify-client-js/blob/71f2937978b1f52d46dc61c418ecec55951edde8/src/resource_clients/actor.ts#L669) parameter with `start` method - even though it would be likely ignored, it might start failing (on type level) now.

Idk, this is probably a fix (so not a breaking change), but IMO still something to consider

barjin (commented):
Thank you @Pijukatel! I have just a few points to discuss:

packages/apify/src/actor.ts
Pijukatel: Yes, I edited PR description to explicitly mention this change. But would you prefer to do this as separate PR with `fix` or is it ok to be included in this one?

docs/guides/log_redirection.mdx
Pijukatel: Formated.

docs/guides/log_redirection.mdx
Pijukatel: Rephrased

packages/apify/src/actor.ts
barjin: Right, that's a good point - we probably should do it in a separate PR, so it shows in the changelog (and is easier to revert if needed).

barjin (approved):
lgtm, thank you for the quick fixes 🚀 

please address the type change in a different PR and feel free to merge 👍

packages/apify/src/actor.ts
Pijukatel: https://github.com/apify/apify-sdk-js/pull/524

PR #524, opened 2025-12-11
fix: Fix type used for `options` argument of `Actor.start`

Pijukatel:
- Use the correct type for `options` argument of `Actor.start` method. Previously used similar type `CallOptions` had a field `waitSecs`, which is not valid in `Actor.start` method.


barjin (approved):
Thanks!

PR #527, opened 2025-12-12
fix: gracefully handle unknown events on the platform

Patai5:
As discussed internally on slacks, adds the missing `?` into the charging manager `calculateMaxEventChargeCountWithinLimit()` method 😎 

The surrounding code suggests that it should've been like this all along :P

And some tests :)

---
### Bug description

On the platform, the charging manager was crashing for unknown events like so:
https://console.apify.com/view/runs/2DP8zdbuYOiQW2nS9

### DIFF
--- a/test/apify/charging.test.ts
+++ b/test/apify/charging.test.ts
@@ -262,31 +262,16 @@ describe('ChargingManager', () => {
             });
 
             test('when running locally should pretend to charge it', async () => {
-                // Arrange
-
-                // Don't use ACTOR_TEST_PAY_PER_EVENT when simulating Apify platform
-                delete process.env.ACTOR_TEST_PAY_PER_EVENT;
-
-                process.env.ACTOR_MAX_TOTAL_CHARGE_USD = '10.0';
-
-                process.env.APIFY_ACTOR_PRICING_INFO = JSON.stringify({
-                    pricingModel: 'PAY_PER_EVENT',
-                    pricingPerEvent: { actorChargeEvents: {} },
-                });
-                process.env.APIFY_CHARGED_ACTOR_EVENT_COUNTS = JSON.stringify(
-                    {},
-                );
+                process.env.ACTOR_TEST_PAY_PER_EVENT = '1';
 
                 await Actor.init();
                 const chargingManager = Actor.getChargingManager();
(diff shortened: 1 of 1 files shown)


test/apify/charging.test.ts
```diff
@@ -209,6 +212,89 @@ describe('ChargingManager', () => {
...
+
+                // Act
+                const chargeResult = await chargingManager.charge({
+                    eventName: 'unknown-event',
+                    count: 5,
+                });
+
+                // Assert
+                expect(chargeResult).toStrictEqual({
+                    chargedCount: 5,
```
Patai5, line 255: Hmm, looking at this retrospectively, this is pretty questionable 🤔 I guess it doesn't really matter, but we didn't really actually charge anything for the case of the unknown event on the platform. Should we return `0` instead?

test/apify/charging.test.ts
```diff
@@ -209,6 +212,89 @@ describe('ChargingManager', () => {
...
+                    count: 3,
+                });
+
+                // Assert
+                expect(chargeResult).toStrictEqual({
+                    chargedCount: 3,
+                    eventChargeLimitReached: false,
+                    chargeableWithinLimit: {},
+                });
+                expect(loggerWarnSpy).not.toHaveBeenCalled();
```
barjin: If testing the local development, why remove `ACTOR_TEST_PAY_PER_EVENT`?

```suggestion
                process.env.ACTOR_TEST_PAY_PER_EVENT = '1';

                await Actor.init();
                const chargingManager = Actor.getChargingManager();

                const chargeResult = await chargingManager.charge({
                    eventName: 'unknown-event',
                    count: 3,
                });

                expect(chargeResult).toStrictEqual({
                    chargedCount: 3,
                    eventChargeLimitReached: false,
                    chargeableWithinLimit: {},
                });
                expect(loggerWarnSpy).not.toHaveBeenCalled();
```

test/apify/charging.test.ts
barjin: I see both reasons for and against it, I say let's keep it as it is :)

barjin (approved):
lgtm, thank you!

PR #528, opened 2025-12-12
fix: Keep track of synthetic apify-default-dataset-item events

### LINKED ISSUE #523: Track the number of `apify-default-dataset-item` events locally
- See https://apify.slack.com/archives/C05GW3FKH17/p1764839874216759 for context
- we should increment the number of `apify-default-dataset-item` events for each push to the default dataset
    - is it enough to handle this in `Actor.pushData`? Or do we need to hook into `Dataset.pushData` as well?
- [ ] if this goes through, make a corresponding issue in Python SDK

janbuchar:
- closes #523

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,7 +4,6 @@ export {
     ChargeResult,
     ActorPricingInfo,
     ChargingManager,
-    mergeChargeResults,
 } from './charging.js';
 export * from './configuration.js';
 export * from './proxy_configuration.js';
--- a/test/apify/charging.test.ts
+++ b/test/apify/charging.test.ts
@@ -4,7 +4,7 @@ import { afterEach, beforeEach, describe, expect, test } from 'vitest';
 
 import log from '@apify/log';
 
-import { mergeChargeResults } from '../../src/index.js';
+import { mergeChargeResults } from '../../src/charging.js';
 import { MemoryStorageEmulator } from '../MemoryStorageEmulator.js';
 
 /**


janbuchar:
@mhamas @metalwarrior665 this proved to be way more complicated than previously thought. I'll try to simplify it on Monday.

packages/apify/src/actor.ts
```diff
@@ -1545,11 +1587,94 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+                    actor.calculatePushDataLimits(items, eventName);
+
+                if (countToCharge > 0) {
+                    await super.pushItems(limitedItems as string[] | Data[]);
+                }
+
+                const chargeResult = await actor.chargingManager.charge({
+                    eventName,
+                    count: countToCharge,
+                });
```
metalwarrior665: We shouldn't be charging platform via API here. So maybe a new option to only increment locally or a new protected method?

metalwarrior665 (commented):
I can see the need for this complexity as we have to handle Dataset.pushData that only connects to SDK via the client

packages/apify/src/actor.ts
```diff
@@ -27,13 +27,14 @@ import { sleep, snakeCaseToCamelCase } from '@crawlee/utils';
...
     ActorStartOptions,
+    ApifyClient,
     ApifyClientOptions,
     RunAbortOptions,
     TaskCallOptions,
     Webhook,
     WebhookEventType,
 } from 'apify-client';
-import { ActorRun as ClientActorRun, ApifyClient } from 'apify-client';
+import { ActorRun as ClientActorRun } from 'apify-client';
```
B4nan: this feels like a type too, if so, it can be a type import (grouped in the line above)

src/charging.ts
```diff
@@ -64,6 +66,23 @@ export interface ActorPricingInfo {
     perEventPrices: Record<string, number>;
 }
 
+export function mergeChargeResults(
```
mhamas, line 69: Suggestion: this is good candidate for unit tests

src/actor.ts
```diff
@@ -1090,37 +1097,29 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+            throw new Error(
+                `Cannot charge for synthetic event '${eventName}' manually`,
+            ); // TODO this should be handled on charging manager level, too (but we also need to track dataset pushes manually 🙈)
         }
 
-        if (eventName) {
-            return await this.chargingManager.charge({
+        const dataset = await this.openDataset();
+
+        if (this.usesPushDataInterception(dataset)) {
```
janbuchar, line 1065: Explain why we need to handle this in two separate code paths

packages/apify/src/actor.ts
```diff
@@ -2315,6 +2318,89 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+        if (context?.chargeResult === undefined) {
+            throw new Error(
+                'Internal error - missing result of charge operation',
+            );
+        }
+
+        return context.chargeResult;
+    }
+
+    private async pushDataViaDirectCharging(
```
janbuchar: The name could be improved here

packages/apify/src/patched_apify_client.ts
```diff
@@ -0,0 +1,130 @@
...
+export const USES_PUSH_DATA_INTERCEPTION = Symbol(
+    'apify:uses-push-data-interception',
+);
+
+export interface PpeAwarePushDataContext {
+    eventName: string | undefined;
+    chargeResult?: ChargeResult;
+}
+
+// Shared context used by Actor.pushData() and PatchedDatasetClient.pushItems().
```
janbuchar: Better explain the purpose of the context (propagate eventName to pushItems calls, gather results of multiple pushItems calls that were triggered by a single pushData call)

packages/apify/src/actor.ts
```diff
@@ -2315,6 +2318,89 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+                eventName: explicitEventName,
+                isDefaultDataset,
+            });
+
+        if (limitedItems.length > 0) {
+            // Preserve original `Dataset.pushData()` call shape for single items.
+            await dataset.pushData(
+                Array.isArray(items) ? limitedItems : limitedItems[0],
+            );
+        }
```
mhamas: Suggestion: Looks duplicated to the intercepted client path. Can this be taken out and unified?

src/charging.ts
```diff
@@ -435,4 +462,55 @@ export class ChargingManager {
 
         return Math.max(0, roundedResult);
     }
+
+    /**
+     * Helper to calculate how many items can be pushed within charging limits.
+     * Returns the limited items and count to charge.
+     */
+    calculatePushDataLimits<T>({
```
mhamas, line 473: Suggestion: I'd add unit tests.

packages/apify/src/actor.ts
barjin: This is not relevant anymore, the signature on L1100 reflects the return type better

packages/apify/src/actor.ts
```diff
@@ -2315,6 +2318,89 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+                    },
+                ),
+            );
+
+            if (explicitEventName !== undefined) {
+                return results[explicitEventName];
+            }
+        }
+
+        return undefined;
```
barjin: `Actor.pushData()` will charge the synthetic event, but will return `undefined` here.

See e.g. the following code, run with `ACTOR_MAX_TOTAL_CHARGE_USD=2`:

```typescript

process.env.ACTOR_TEST_PAY_PER_EVENT = '1';

await Actor.init();
console.log(await Actor.pushData([{ 'a': 1 }, { 'b': 2 }, { 'c': 3 }])); // prints undefined
console.log(await Actor.pushData([{ 'a': 1 }, { 'b': 2 }, { 'c': 3 }])); // prints undefined
console.log(await Actor.pushData([{ 'a': 1 }, { 'b': 2 }, { 'c': 3 }])); // prints undefined
console.log(await Actor.pushData([{ 'a': 1 }, { 'b': 2 }, { 'c': 3 }])); // prints undefined

await Actor.exit();
```

This will run all the calls, but will store only **two results** into the dataset. The developer has no way of returning early, as they won't get the charging manager response here.

packages/apify/src/actor.ts
```diff
@@ -2315,6 +2318,89 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
+
+        // Always return `ChargeResult` when user explicitly asked to charge for an event.
+        // This includes the case where the count is 0 (e.g. budget exhausted), because
+        // `ChargingManager.charge({ count: 0 })` is a cheap no-op and does not call the API.
+        if (Object.keys(eventsToCharge).length > 0) {
+            const results: Record<string, ChargeResult> = {};
+            await Promise.all(
+                Object.entries(eventsToCharge).map(
+                    async ([eventName, count]) => {
+                        results[eventName] = await this.chargingManager.charge({
```
barjin: calculatePushDataLimits will contain the apify-default-dataset-item key (see [here](https://github.com/apify/apify-sdk-js/blob/4477dde487805724e728b6d5a7fc7a56e38b0458/packages/apify/src/charging.ts#L506)). Here we'll manually charge the synthetic event. Is this expected?

packages/apify/src/index.ts
```diff
@@ -1,5 +1,11 @@
 export * from './actor.js';
-export * from './charging.js';
+export {
+    ChargeOptions,
+    ChargeResult,
+    ActorPricingInfo,
+    ChargingManager,
+    mergeChargeResults,
```
barjin: Should we export `mergeChargeResults` to outside SDK users? The relevant test is using relative import anyway.

src/charging.ts
```diff
@@ -435,4 +465,114 @@ export class ChargingManager {
...
+    const { limitedItems, eventsToCharge } =
+        chargingManager.calculatePushDataLimits({
+            items,
+            eventName,
+            isDefaultDataset,
+        });
+
+    if (limitedItems.length > 0) {
+        // Preserve original call shape for single items
+        await pushFn(
```
barjin, line 552: This `await` causes race conditions on concurrent calls to `pushDataAndCharge`, as the pushed items haven't been charged for yet.

```typescript
import { pushDataAndCharge } from './packages/apify/src/charging.ts';
import { Actor } from './packages/apify/src/index.ts';

const dataset: any[] = [];

async function main() {
    await Actor.init();

    async function pushData(data: string, count: number) {
        return pushDataAndCharge({
            chargingManager: Actor.getChargingManager(),
            eventName: 'a-pricy-one',
            isDefaultDataset: false,
            items: Array.from({ length: count }, () => ({ data })),
            pushFn: async (data) => { dataset.push(...(data as any)); }
        });
    }

    // awaiting in sequence works correctly
    await pushData('A', 10);
    await pushData('B', 10);

    // concurrent push stores 2x as many items, as we're only pushing 
    // A items (having not charged them yet) when calculating B limits
    await Promise.all([
        pushData('A', 10),
        pushData('B', 10)
    ]);

    console.log(dataset);

    await Actor.exit();
}

main();
```

Running e.g. `ACTOR_MAX_TOTAL_CHARGE_USD=2 ACTOR_TEST_PAY_PER_EVENT=1 npx tsx ./test.ts` will return 2 results (A A) with sequential `await`, but 4 results (A A B B) for the `Promise.all` call

janbuchar: Yes, if I understand you correctly, this has been an issue before this PR. We could add some kind of a mutex later 🤷

metalwarrior665: There was discussion about this. I think the output was that this is not a big deal and we will err on the side of giving free data.

janbuchar: On the other hand, in Python, which has mutexes in the standard library, we just wrapped this routine with one.

PR #535, opened 2025-12-19
chore(ci): Set up trusted NPM publishing

### LINKED ISSUE #437: Set up OIDC npm publishing
https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/

janbuchar:
- closes #437
- I did not remove lerna completely, but it doesn't make github tags and releases now because it was really weird how it made two commits per release


.github/workflows/publish-to-npm.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+            - name: Publish package (latest)
+              if: ${{ inputs.tag == 'latest' }}
+              run: |
+                  npx lerna publish from-package --contents dist --yes
+
+            - name: Publish package (beta)
+              if: ${{ inputs.tag == 'beta' }}
+              run: |
+                  npx lerna publish --canary --preid beta --dist-tag next --yes
```
janbuchar, line 67: We are pushing beta releases to apify@next here - is that intentional?

barjin: I believe so, we do the same thing (`next` tag for beta) with Crawlee ([npm](https://www.npmjs.com/package/crawlee?activeTab=versions)).

vladfrangu: Don't we use `next` everywhere?

barjin: E.g. apify-client has `beta` ([npm](https://www.npmjs.com/package/apify-client?activeTab=versions)), not sure what's the reason though (probably legacy).

barjin (approved):
I didn't find anything obvious, it's a go from me 👍 

Thank you!

janbuchar:
> Is there any reason this repo still uses lerna? We could move to just `yarn npm publish` in the future unless we want monorepo to stay

Nah, no reason except I didn't want to change too many things at once

B4nan (approved):
i'd just rename the workflow to use dashes as discussed on slack. 

its a mess, every repo does it differently, but here this would be the only workflow that uses dashes.

PR #539, opened 2026-01-08
chore: Remove the monorepo setup

### LINKED ISSUE #533: Remove the monorepo setup
After removing scraper-tools and generic scrapers, we can make this a single-package repository

nikitachapovskii-dev:
- closes #533

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,6 @@
 [![Chat on discord](https://img.shields.io/discord/801163717915574323?label=discord)](https://discord.gg/jyEM2PRvMU)
 [![Build Status](https://github.com/apify/apify-sdk-js/actions/workflows/test-and-release.yaml/badge.svg?branch=master)](https://github.com/apify/apify-sdk-js/actions/workflows/test-and-release.yaml)
 
-## Upgrading from v2
-
-Visit the [Upgrading Guide](https://docs.apify.com/sdk/js/docs/upgrading/upgrading-to-v3) to find out what changes you need to make (especially the section related to this very [Apify SDK](https://docs.apify.com/sdk/js/docs/upgrading/upgrading-to-v3#apify-sdk)), and, if you encounter any issues, join our [Discord server](https://discord.gg/jyEM2PRvMU) for help!
-
 ## Quick Start
 
 This short tutorial will set you up to start using Apify SDK in a minute or two.
@@ -43,6 +39,10 @@ await Actor.exit();
 
 If you find any bug or issue with the Apify SDK, please [submit an issue on GitHub](https://github.com/apify/apify-sdk-js/issues).
 For questions, you can ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/apify) or contact support@apify.com
+
+## Upgrading
+
+Visit the [Upgrading Guide](https://docs.apify.com/sdk/js/docs/upgrading) to find out what changes you might want to make, and, if you encounter any issues, join our [Discord server](https://discord.gg/jyEM2PRvMU) for help!
 
 ## Contributing
 


.github/workflows/release.yaml
```diff
@@ -94,40 +85,23 @@ jobs:
...
+                  npm version ${{ inputs.version }} --no-git-tag-version
 
             - name: Get the updated package version
               id: updated-version
               run: |
-                  UPDATED_VERSION=$(node -p "require('./packages/apify/package.json').version")
+                  UPDATED_VERSION=$(node -p "require('./package.json').version")
                   echo result=$UPDATED_VERSION | tee $GITHUB_OUTPUT
 
             - name: Extract latest CHANGELOG entry
```
janbuchar: This will now do nothing - the changes in the changelog were done by lerna and this PR removes that. 

You have two options:

1. Utilize the [git-cliff-release Action](https://github.com/apify/workflows/tree/main/git-cliff-release) - you can check out [apify-client-js](https://github.com/apify/apify-client-js/) to see how it's done
2. Recreate what lerna does using conventional-changelog (not 100% sure about the exact package)

What do you think? Opinions, @B4nan @barjin?

CODEOWNERS
```diff
@@ -1,5 +1,7 @@
 # Code owners are automatically requested for review when someone opens a pull request that modifies code that they own.
+
 # See: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
 
 # To keep eye on new functions added to Actor specification
-/packages/apify/src/            @jancurn @mnmkng
+
+@jancurn @mnmkng
```
janbuchar: I believe we can just remove this entry. No need to tag two C-levels because of every change.

.github/workflows/release.yaml
B4nan: 1. is correct, we dont want yet another custom setup here

website/docusaurus.config.js
```diff
@@ -134,7 +134,7 @@ module.exports = {
                     );
                 },
                 packages: packages.map((name) => ({
-                    path: `packages/${name}`,
+                    path: `.`,
```
B4nan: we might want to clean this up a bit too, that plugin has a single package mode

test/apify/charging.test.ts
```diff
@@ -3,7 +3,7 @@ import { afterEach, beforeEach, describe, expect, test } from 'vitest';
 
 import log from '@apify/log';
 
-import { Actor } from '../../packages/apify/src/index.js';
+import { Actor } from '../../src/index.js';
```
B4nan:
```suggestion
import { Actor } from 'apify';
```

.github/workflows/publish-to-npm.yaml
```diff
@@ -58,12 +49,20 @@ jobs:
...
+                  BASE_VERSION=$(node -p "require('./package.json').version")
+                  BETA_VERSION="${BASE_VERSION}-beta.${GITHUB_RUN_NUMBER}"
+                  npm version "$BETA_VERSION" --no-git-tag-version
+                  node -p "require('./package.json').version"
+
             - name: Publish package (latest)
               if: ${{ inputs.tag == 'latest' }}
               run: |
-                  npx lerna publish from-package --contents dist --yes
+                  npm publish --tag latest --access public --provenance
```
B4nan: `--tag latest` is the default and i dont think we need `--access public` anywhere, that's for publishing namespaced packages like `@apify/...`

barjin: Same for the `--provenance` flag with OIDC publishing, see [docs](https://docs.npmjs.com/trusted-publishers#automatic-provenance-generation).

vitest.config.mts
barjin: nit - this can likely go too

CODEOWNERS
barjin: we can remove the entire file, this was the only rule

README.md
```diff
@@ -12,9 +12,9 @@ This monorepo also used to hold the source code of generic scraper Actors. These
 
 > Would you like to work with us on Crawlee, Apify SDK or similar projects? [We are hiring!](https://apify.com/jobs#senior-node.js-engineer)
 
-| package                               | version                                                                                       |
-| ------------------------------------- | --------------------------------------------------------------------------------------------- |
-| [`apify`](./packages/apify/README.md) | [![NPM version](https://img.shields.io/npm/v/apify.svg)](https://www.npmjs.com/package/apify) |
+| package                                                                | version                                                                                       |
+| ---------------------------------------------------------------------- | --------------------------------------------------------------------------------------------- |
+| [`apify`](https://github.com/apify/apify-sdk-js/blob/master/README.md) | [![NPM version](https://img.shields.io/npm/v/apify.svg)](https://www.npmjs.com/package/apify) |
```
barjin: As disucssed yesterday, the top-level readme doesn't make much sense anymore (see e.g. this table). 

How about we move the `apify` package README.md to the top-level of the repo?

README.md
```diff
@@ -1,25 +1,79 @@
...
+# Apify SDK
 
 [![npm version](https://badge.fury.io/js/apify.svg)](https://www.npmjs.com/package/apify)
 [![Downloads](https://img.shields.io/npm/dm/apify.svg)](https://www.npmjs.com/package/apify)
 [![Chat on discord](https://img.shields.io/discord/801163717915574323?label=discord)](https://discord.gg/jyEM2PRvMU)
 [![Build Status](https://github.com/apify/apify-sdk-js/actions/workflows/test-and-release.yaml/badge.svg?branch=master)](https://github.com/apify/apify-sdk-js/actions/workflows/test-and-release.yaml)
 
-Apify SDK is the core set of tools and utilities that we've built to help make your interaction with the [Apify Platform](https://apify.com) easier.
-This monorepo holds all the components and tools that we've created for it!
+Apify SDK provides the tools required to run your own Apify Actors. The crawlers and scraping related tools, previously included in Apify SDK (v2), have been split into a brand-new module - [`crawlee`](https://npmjs.org/crawlee), while keeping the Apify specific parts in this module.
```
janbuchar: All this v2-related stuff is super dated, maybe we could reword or completely remove it?

README.md
```diff
@@ -1,25 +1,79 @@
...
 
-This monorepo also used to hold the source code of generic scraper Actors. These have been moved to a separate repository - [apify/actor-scraper](https://github.com/apify/actor-scraper).
+> Would you like to work with us on Crawlee, Apify SDK or similar projects? We are hiring [Node.js engineers](https://apify.com/jobs#senior-node.js-engineer).
```
janbuchar: nit - the anchor is not valid anymore, so this leads to the general careers page. I'm not sure if that's a problem, but it'd be better to either update the link or completely remove it.

.github/workflows/publish-to-npm.yaml
```diff
@@ -58,12 +49,20 @@ jobs:
             - name: Build module
               run: npm run build
 
+            - name: Set beta version
+              if: ${{ inputs.tag == 'beta' }}
+              run: |
+                  BASE_VERSION=$(node -p "require('./package.json').version")
+                  BETA_VERSION="${BASE_VERSION}-beta.${GITHUB_RUN_NUMBER}"
```
janbuchar: This is different from the original beta version scheme which used the number of commits to master since the last tag. It is also different from most other repositories that increment the beta number manually to maintain a gapless sequence.

Here, we can:
1. stick to the original scheme (number of commits since last stable tag)
2. use this brand new scheme based on run number
3. migrate to the "custom gapless sequence" approach
4. think of something completely different
5. 2, 3 or 4 and propagate the change to all other repositories

Personally, I prefer 1 or 3 just to drive this PR home and then 5+4 as a separate initiative.

CC @B4nan @barjin @vladfrangu

.github/workflows/test-and-release.yaml
```diff
@@ -148,28 +120,9 @@ jobs:
...
-                  echo "changed_packages=$(npx lerna changed -p | wc -l | xargs)" >> "$GITHUB_OUTPUT"
-
-            - name: Check for changes 2
-              if: steps.changed_packages.outputs.changed_packages == '0'
-              run: |
-                  echo "no changes detected by lerna"
-
             - name: Publish beta to npm
               uses: apify/workflows/execute-workflow@main
               if: steps.changed_packages.outputs.changed_packages != '0'
```
janbuchar: This condition has to go, too

janbuchar: We should also remove this

README.md
```diff
@@ -1,25 +1,79 @@
...
+
+Apify SDK requires [Node.js](https://nodejs.org/en/) 16 or later. Add Apify SDK to any Node.js project by running:
+
+```bash
+npm install apify crawlee playwright
+```
+
+> For this example, we'll also install the [`crawlee`](https://npmjs.org/crawlee) module, as it now provides the crawlers that were previously exported by Apify SDK. If you don't plan to use crawlers in your Actors, then you don't need to install it. Keep in mind that neither `playwright` nor `puppeteer` are bundled with `crawlee` in order to reduce install size and allow greater flexibility. That's why we manually install it with NPM. You can choose one, both, or neither.
+
+There are two ways to initialize your Actor: by using the `Actor.main()` function you're probably used to, or by calling `Actor.init()` and `Actor.exit()` manually. We prefer explicitly calling `init` and `exit`.
```
janbuchar: Is one of these preferred over the other? I'd say that we almost exclusively use `Actor.init` - might as well just mention that in the README so that we don't confuse users..

README.md
```diff
@@ -1,25 +1,79 @@
...
+const crawler = new PlaywrightCrawler({
+    async requestHandler({ request, page, enqueueLinks }) {
+        // Extract HTML title of the page.
+        const title = await page.title();
+        console.log(`Title of ${request.url}: ${title}`);
+
+        // Add URLs that point to the same hostname.
+        await enqueueLinks();
+    },
+});
```
janbuchar: Maybe the quickstart could use a more simplistic example without crawlee - e.g., just push something and exit. Then we could link to the "Run crawlers on Apify" docs page for crawlee for details on how to run crawlee stuff on Apify.

.github/workflows/publish-to-npm.yaml
B4nan: > migrate to the "custom gapless sequence" approach

I'd prefer this one. Btw are you sure the run numbers are guaranteed to be sequential? Gaps would be ok, but the versions need to be "growing".

barjin: > 1 or 3 just to drive this PR home and then 5+4 as a separate initiative.

Yes from me, let's just clone the `apify-client-js` approach to be consistent and forget this :)

README.md
B4nan: yeah lets do that, with top level await the `main` method is needed only for a very specific cases with multiple crawlers in one context

.github/scripts/before-beta-release.js
janbuchar, line 1: Did you make any changes to the file or is it a direct copy of the one in the JS client?

.github/workflows/test-and-release.yaml
```diff
@@ -148,35 +120,5 @@ jobs:
...
-            - name: Publish beta to npm
-              uses: apify/workflows/execute-workflow@main
-              if: steps.changed_packages.outputs.changed_packages != '0'
-              with:
-                  workflow: publish-to-npm.yaml
-                  inputs: >
-                      {
-                          "ref": "${{ github.sha }}",
-                          "tag": "beta"
-                      }
```
janbuchar, line 182: I think we shouldn't remove this block, we should just execute it unconditionally

package.json
janbuchar, line 1: Now we still have duplicates of `package.json` and co. under packages/apify - can you move or remove those?

package.json
```diff
@@ -1,10 +1,21 @@
 {
     "name": "apify-sdk-js",
-    "private": true,
-    "description": "Apify SDK monorepo",
-    "workspaces": [
-        "packages/apify"
-    ],
+    "version": "3.5.3",
+    "description": "Apify SDK",
```
janbuchar: Please copy the description from the other package.json

.github/scripts/before-beta-release.js
nikitachapovskii-dev: that is a direct copy paste from script in JS client repo.

package.json
nikitachapovskii-dev: I think on this stage we may want to completely delete the folder packages/apify with all the content inside.

janbuchar: Probably, yes. Just make sure that you don't throw out the baby with the bathwater :slightly_smiling_face:

README.md
```diff
@@ -1,25 +1,60 @@
...
+If you want to learn more, proceed to the [Apify Platform](https://docs.apify.com/sdk/js/docs/guides/apify-platform)
+guide that will take you step by step through running your Actor on Apify's platform.
 
-## Apify SDK
+Apify SDK requires [Node.js](https://nodejs.org/en/) 16 or later. Add Apify SDK to any Node.js project by running:
 
-Apify SDK provides the tools required to run your own Apify Actors! The crawlers and scraping related tools, previously included in Apify SDK (v2), have been split into
-a brand-new module - [`crawlee`](https://npmjs.org/crawlee) (which you can use outside Apify too!), while keeping the Apify specific parts in this module!
+```bash
+npm install apify crawlee playwright
```
janbuchar: No need to install all this, just `apify` should do it.

README.md
janbuchar: @nikitachapovskii-dev this is not resolved yet

README.md
```diff
@@ -1,25 +1,60 @@
...
 
-Apify SDK is the core set of tools and utilities that we've built to help make your interaction with the [Apify Platform](https://apify.com) easier.
-This monorepo holds all the components and tools that we've created for it!
+## Upgrading from v2
 
-This monorepo also used to hold the source code of generic scraper Actors. These have been moved to a separate repository - [apify/actor-scraper](https://github.com/apify/actor-scraper).
+A lot of things have changed since version 2 of the Apify SDK, including the split of the crawlers to the new [`crawlee`](https://npmjs.org/crawlee) module. We've written a guide to help you easily migrate from v2 to v3. Visit the [Upgrading Guide](https://docs.apify.com/sdk/js/docs/upgrading/upgrading-to-v3) to find out what changes you need to make (especially the section related to this very [Apify SDK](https://docs.apify.com/sdk/js/docs/upgrading/upgrading-to-v3#apify-sdk)), and, if you encounter any issues, join our [Discord server](https://discord.gg/jyEM2PRvMU) for help!
```
janbuchar:
```suggestion
Visit the [Upgrading Guide](https://docs.apify.com/sdk/js/docs/upgrading/upgrading-to-v3) to find out what changes you need to make (especially the section related to this very [Apify SDK](https://docs.apify.com/sdk/js/docs/upgrading/upgrading-to-v3#apify-sdk)), and, if you encounter any issues, join our [Discord server](https://discord.gg/jyEM2PRvMU) for help!
```

README.md
janbuchar: ...but it wasn't necessary to remove the whole command

README.md
```diff
@@ -1,25 +1,45 @@
...
 
-Apify SDK provides the tools required to run your own Apify Actors! The crawlers and scraping related tools, previously included in Apify SDK (v2), have been split into
-a brand-new module - [`crawlee`](https://npmjs.org/crawlee) (which you can use outside Apify too!), while keeping the Apify specific parts in this module!
+```bash
+npm install apify
+```
+
+To initialize your Actor and to stop it use the `Actor.init()` and `Actor.exit()` functions. You also may use `Actor.main()` function for cases with multiple crawlers in one context.
+
+### Using `Actor.init()` and `Actor.exit()`
```
janbuchar:
```suggestion
```

package.json
```diff
@@ -1,10 +1,21 @@
 {
     "name": "apify-sdk-js",
```
janbuchar: This needs to be changed to just `apify`

package.json
```diff
@@ -32,16 +43,22 @@
         "url": "https://github.com/apify/apify-sdk-js/issues"
     },
     "homepage": "https://docs.apify.com/sdk/js",
+    "publishConfig": {
+        "access": "public"
+    },
```
janbuchar: I believe this is not necessary

package.json
```diff
@@ -32,16 +43,22 @@
...
-        "clean": "turbo run clean",
-        "build": "turbo run build",
-        "ci:build": "turbo run build --cache-dir=\".turbo\"",
+        "clean": "rimraf ./dist",
+        "compile": "tsc -p tsconfig.build.json && gen-esm-wrapper ./dist/index.js ./dist/index.mjs",
+        "copy": "tsx ./scripts/copy.ts --readme=local",
+        "fixApifyExport": "node ./scripts/temp_fix_apify_exports.mjs",
+        "build": "npm run clean && npm run compile && npm run copy && npm run fixApifyExport",
+        "prepublishOnly": "npm run build",
+        "ci:build": "npm run build",
```
janbuchar, line 54: I suspect these two are obsolete - @B4nan @vladfrangu?

package.json
B4nan: indeed, it was there because of the scraper tools

package.json
B4nan: no idea, but `ci:build` is still used in the workflows. i can imagine there was a reason for this, but it could be easily lerna specific

janbuchar (approved):

README.md
```diff
@@ -1,25 +1,43 @@
...
+# Apify SDK
 
 [![npm version](https://badge.fury.io/js/apify.svg)](https://www.npmjs.com/package/apify)
 [![Downloads](https://img.shields.io/npm/dm/apify.svg)](https://www.npmjs.com/package/apify)
 [![Chat on discord](https://img.shields.io/discord/801163717915574323?label=discord)](https://discord.gg/jyEM2PRvMU)
 [![Build Status](https://github.com/apify/apify-sdk-js/actions/workflows/test-and-release.yaml/badge.svg?branch=master)](https://github.com/apify/apify-sdk-js/actions/workflows/test-and-release.yaml)
 
-Apify SDK is the core set of tools and utilities that we've built to help make your interaction with the [Apify Platform](https://apify.com) easier.
-This monorepo holds all the components and tools that we've created for it!
+## Upgrading from v2
```
B4nan: i wouldnt put this as the very first thing in the readme, v3 has been around for more than 3 years now, i dont think we need to mention v2 at all. i would keep the link to upgrading guides at the end of the readme (lets use this link: https://docs.apify.com/sdk/js/docs/upgrading)

B4nan (approved):
just one nit

README.md
nikitachapovskii-dev: is it better to put it to the bottom of the file or after, let's say, **support** section?

B4nan: somewhere at the bottom, not necessarily very bottom, i'll leave that up to you

PR #548, opened 2026-02-03
chore: Remove monorepo setup - drop files from published packages

### LINKED ISSUE #547: E2E tests fail because of large package size
Looks like [removing the monorepo setup from JS SDK](https://github.com/apify/apify-sdk-js/pull/539) broke the [scheduled e2e SDK tests](https://github.com/apify/apify-sdk-js/actions/runs/21087102645/job/60652325933), probably because it made the JS bundle bigger.

nikitachapovskii-dev:
The PR fixes https://github.com/apify/apify-sdk-js/pull/539 which broke the scheduled e2e SDK tests

Closes #547

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
     },
     "homepage": "https://docs.apify.com/sdk/js",
     "files": [
-        "dist"
+        "dist",
+        "!dist/tsconfig.build.tsbuildinfo"
     ],
     "scripts": {
         "prepare": "husky",


barjin (changes requested):
This doesn't seem to fix the linked issue - check out the output of `npm publish --dry-run` (and compare with the contents of the [Code tab](https://www.npmjs.com/package/apify?activeTab=code) for the last `latest` release).

Also, check out some tips on `.npmignore` [in the docs](https://docs.npmjs.com/cli/v11/using-npm/developers#keeping-files-out-of-your-package) - e.g. `node_modules` is always ignored etc.

And please update the PR title to something more descriptive :)

.npmignore
janbuchar, line 1: I think we could disable everything on the first line and then just allow dist/ and possibly package.json (now it's included twice and I'm not sure which one is actually used - I guess the root one?).

This way, every time we add any file to the root, it will be included in the package and we most likely don't want that.

barjin: See the `files` field in `package.json` ([docs](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#files)), which leads to imo a cleaner setup.

Regarding the double `package.json` (and `README.md`) - I'm pretty positive we're copying it to `dist/` at some point, maybe we can drop that script first

B4nan: We should be publishing only the contents of `dist` folder. That's how things worked with the lerna setup before, the copy script was initially about copying readme and package.json into the dist folder, so we can run npm publish inside it, that's why its called `copy` in the first place.

barjin: Copying makes sense to me in monorepos, but for a single-package project like this, it IMO only leads to a more complicated setup. npm publishes the root `package.json` and `README.md` by default, we can just add `dist`, set the right paths in `package.json`, and be done (with no `postBuild` scripts). 

See e.g. `apify-client` setup [here](https://www.npmjs.com/package/apify-client?activeTab=code), it does the same thing.

B4nan: Maybe it's just my personal thing, but I hate publishing `dist` folder. Monorepo or not.

janbuchar: No strong opinion here, but let's minimize junk in the package and let's not include new stuff in the root into the package by default

nikitachapovskii-dev: moved ignoring to package.json, please see the result:

npm notice 📦  apify@3.5.3-dryrun.1770194821
npm notice Tarball Contents
npm notice 10.9kB LICENSE.md
npm notice 3.2kB README.md
npm notice 57.7kB dist/actor.d.ts
npm notice 11.3kB dist/actor.d.ts.map
npm notice 71.9kB dist/actor.js
npm notice 25.3kB dist/actor.js.map
npm notice 5.3kB dist/charging.d.ts
npm notice 2.4kB dist/charging.d.ts.map
npm notice 17.8kB dist/charging.js
npm notice 11.1kB dist/charging.js.map
npm notice 7.7kB dist/configuration.d.ts
npm notice 1.8kB dist/configuration.d.ts.map
npm notice 9.8kB dist/configuration.js
npm notice 3.4kB dist/configuration.js.map
npm notice 783B dist/index.d.ts
npm notice 625B dist/index.d.ts.map
npm notice 1.9kB dist/index.js
npm notice 422B dist/index.js.map
npm notice 731B dist/index.mjs
npm notice 381B dist/input-schemas.d.ts
npm notice 324B dist/input-schemas.d.ts.map
npm notice 2.7kB dist/input-schemas.js
npm notice 1.9kB dist/input-schemas.js.map
npm notice 579B dist/key_value_store.d.ts
npm notice 431B dist/key_value_store.d.ts.map
npm notice 1.7kB dist/key_value_store.js
npm notice 1.1kB dist/key_value_store.js.map
npm notice 1.3kB dist/patched_apify_client.d.ts
npm notice 697B dist/patched_apify_client.d.ts.map
npm notice 2.5kB dist/patched_apify_client.js
npm notice 1.6kB dist/patched_apify_client.js.map
npm notice 3.0kB dist/platform_event_manager.d.ts
npm notice 447B dist/platform_event_manager.d.ts.map
npm notice 5.6kB dist/platform_event_manager.js
npm notice 2.1kB dist/platform_event_manager.js.map
npm notice 11.3kB dist/proxy_configuration.d.ts
npm notice 2.2kB dist/proxy_configuration.d.ts.map
npm notice 16.6kB dist/proxy_configuration.js
npm notice 7.9kB dist/proxy_configuration.js.map
npm notice 120.6kB dist/tsconfig.build.tsbuildinfo
npm notice 365B dist/utils.d.ts
npm notice 180B dist/utils.d.ts.map
npm notice 2.8kB dist/utils.js
npm notice 1.5kB dist/utils.js.map
npm notice 4.1kB package.json
npm notice Tarball Details
npm notice name: apify
npm notice version: 3.5.3-dryrun.1770194821
npm notice filename: apify-3.5.3-dryrun.1770194821.tgz
npm notice package size: 105.2 kB
npm notice unpacked size: 448.7 kB
npm notice shasum: 18931646a5016df6f75392a163d0deb23e6fd57a
npm notice integrity: sha512-eYjtk1IFt9aTq[...]f6HMPf3gdjl3Q==
npm notice total files: 46

janbuchar: I guess you can remove this file now, right?

package.json
```diff
@@ -43,6 +43,12 @@
         "url": "https://github.com/apify/apify-sdk-js/issues"
     },
     "homepage": "https://docs.apify.com/sdk/js",
+    "files": [
+        "dist",
+        "!dist/README.md",
+        "!dist/package.json",
+        "!dist/LICENSE.md"
+    ],
```
janbuchar, line 49: These are put there by `scripts/copy.ts` - we can just stop doing that and remove these exclusions.

scripts/copy.ts
```diff
@@ -27,21 +27,6 @@ function rewrite(path: string, replacer: (from: string) => string): void {
...
-        break;
-    }
-    default: {
-        copy('README.md', root, target);
-    }
-}
-
-copy('LICENSE.md', root, target);
-copy('package.json', process.cwd(), target);
 rewrite(resolve(target, 'package.json'), (pkg) => {
```
janbuchar: What is this supposed to do? If it's not necessary, we could remove the script completely.

nikitachapovskii-dev: Yeah, assuming that we removed copy, rewrite will do nothing.
On my way to delete the whole script.

barjin (commented):
Mostly lgtm, thank you! 

As the last step, please drop the `.tsbuildinfo` file from the `dist` folder (it's almost 25% of the unpacked size). You can either update the `"files"` key, or shuffle around the tsconfigs a little (see more [here](https://github.com/apify/crawlee/pull/3243)).

PR #583, opened 2026-03-25
refactor: adapt SDK Configuration for crawlee Configuration redesign

B4nan:
## Summary
- Adapts the SDK's `Configuration` class extension to work with the new field-based Configuration from crawlee (apify/crawlee#3080)
- Replaces `.get('key')` / `.set('key', value)` calls with direct property access (`.key`) across all SDK files
- Replaces `config.useStorageClient()` / `config.useEventManager()` with `serviceLocator.setStorageClient()` / `serviceLocator.setEventManager()`
- Adds `zod` as a direct dependency (SDK defines its own config fields using zod schemas)
- Fixes tests that relied on mutable configuration (env var mutation after init)

## Dependencies
- Depends on apify/crawlee#3080 being merged and released first
- There are also unrelated v4 API changes in crawlee core (EventManager constructor, ProxyConfiguration signatures, etc.) that will need separate fixes before this fully type-checks


packages/apify/src/proxy_configuration.ts
```diff
@@ -438,7 +438,7 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
      */
     // TODO: Make this private
     protected async _setPasswordIfToken(): Promise<void> {
-        const token = this.config.get('token');
+        const {token} = (this.config as Configuration);
```
barjin: Why are these casts necessary?

packages/apify/src/platform_event_manager.ts
```diff
@@ -62,7 +62,7 @@ export class PlatformEventManager extends EventManager {
         }
 
         await super.init();
-        const eventsWsUrl = this.config.get('actorEventsWsUrl');
+        const eventsWsUrl = (this.config as Configuration).actorEventsWsUrl;
```
barjin: Same here, `this.config` should be a `Configuration`, no?

packages/apify/src/configuration.ts
```diff
@@ -1,45 +1,206 @@
...
+    isAtHome: field(coerceBoolean.optional(), 'APIFY_IS_AT_HOME'),
+    userId: field(z.string().optional(), 'APIFY_USER_ID'),
+    inputSecretsPrivateKeyPassphrase: field(
+        z.string().optional(),
+        'APIFY_INPUT_SECRETS_PRIVATE_KEY_PASSPHRASE',
+    ),
+    inputSecretsPrivateKeyFile: field(
+        z.string().optional(),
+        'APIFY_INPUT_SECRETS_PRIVATE_KEY_FILE',
+    ),
```
barjin: nit: many of these envvar names exist in the `@apify/consts` `APIFY_ENV_VARS` export ([see here](https://github.com/apify/apify-shared-js/blob/7a0ab95875eb5f1d00fdb679a08e988597cb54a6/packages/consts/src/consts.ts#L282))

packages/apify/src/charging.ts
```diff
@@ -87,20 +87,19 @@ export class ChargingManager {
...
 
     constructor(configuration: Configuration, apifyClient: ApifyClient) {
-        this.maxTotalChargeUsd =
-            configuration.get('maxTotalChargeUsd') || Infinity; // convert `0` to `Infinity` in case the value is an empty string
-        this.isAtHome = configuration.get('isAtHome');
-        this.actorRunId = configuration.get('actorRunId');
-        this.purgeChargingLogDataset = configuration.get('purgeOnStart');
-        this.useChargingLogDataset = configuration.get('useChargingLogDataset');
+        this.maxTotalChargeUsd = configuration.maxTotalChargeUsd || Infinity; // convert `0` to `Infinity` in case the value is an empty string
+        this.isAtHome = !!configuration.isAtHome;
```
barjin: Should this (`|| Infinity`) happen in `Configuration`'s `coerceNumber` directly (e.g. treat empty string as missing value rather than `0`)? 

Additionally, isn't `configuration.isAtHome` a `Boolean` at this point? (regarding the `!!`)

barjin (approved):
Thank you @B4nan , I just have a few questions / refactoring nits, nothing major 👍

packages/apify/src/charging.ts
B4nan: Crawlee Core v4's `Configuration` already treats empty-string env vars as unset at the schema layer, so `ACTOR_MAX_TOTAL_CHARGE_USD=""` resolves to `undefined` (not `0`) — no custom preprocessor needed in our `coerceNumber`. The remaining `|| Infinity` here only handles explicit-`0` input, where `0` means "no limit"; tightened the inline comment to reflect that.

On the `!!isAtHome` part — switched to `?? false`. `configuration.isAtHome` is `boolean | undefined` (`coerceBoolean.optional()`), so the two are equivalent.

PR #586, opened 2026-04-14
docs: add status message section to Actor lifecycle page

TC-MO:
I was going through apify-docs issues and noticed https://github.com/apify/apify-docs/issues/400.

This PR adds similar section as is already in Python docs & creates a new codeblock demonstrating the functionality.

### DIFF
--- a/docs/02_concepts/01_actor_lifecycle.mdx
+++ b/docs/02_concepts/01_actor_lifecycle.mdx
@@ -13,6 +13,7 @@ import CodeBlock from '@theme/CodeBlock';
 
 import MainSource from '!!raw-loader!./apify_platform_main.ts';
 import InitExitSource from '!!raw-loader!./apify_platform_init_exit.ts';
+import StatusMessageSource from '!!raw-loader!./code/status_message.ts';
 
 [Apify](https://apify.com) is a platform built to serve large-scale and high-performance web scraping
 and automation needs. It provides easy access to compute instances (_Actors_),
@@ -197,6 +198,18 @@ When the <ApiLink to="apify/class/Dataset">`Dataset`</ApiLink> is stored on the 
 - [Key-value stores API reference](https://apify.com/docs/api/v2#/reference/key-value-stores)
 - [Datasets API reference](https://docs.apify.com/api/v2#/reference/datasets)
 - [Request queues API reference](https://docs.apify.com/api/v2#/reference/request-queues)
+
+## Status messages
+
+Status messages are human-readable messages that provide information about an Actor's progress. Unlike logs, status messages are visible at a glance in the Apify Console on the Actor run details page, making them useful for monitoring long-running Actors without digging through logs.
+
+Use [`Actor.setStatusMessage()`](https://sdk.apify.com/api/apify/class/Actor#setStatusMessage) to update the status message at meaningful milestones during your Actor's execution. Avoid updating it on every single item — focus on key transitions like starting a new phase, reaching a progress milestone, or completing the run.
+
+<CodeBlock language="js">{StatusMessageSource}</CodeBlock>
+
+The SDK only sends an API request when the status message actually changes, so you don't need to worry about duplicate updates.
+
+When your Actor finishes, you can mark the final status message as **terminal** using the `isStatusMessageTerminal` option. A terminal status message persists after the run completes, so users can see the final result at a glance.
 
 ## Environment variables
 
--- a/docs/02_concepts/01_actor_lifecycle.mdx
+++ b/docs/02_concepts/01_actor_lifecycle.mdx
@@ -201,7 +201,7 @@ When the <ApiLink to="apify/class/Dataset">`Dataset`</ApiLink> is stored on the 
 
 ## Status messages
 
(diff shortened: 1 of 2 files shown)


docs/02_concepts/01_actor_lifecycle.mdx
```diff
@@ -242,6 +199,18 @@ When the <ApiLink to="apify/class/Dataset">`Dataset`</ApiLink> is stored on the
 - [Datasets API reference](https://docs.apify.com/api/v2#/reference/datasets)
 - [Request queues API reference](https://docs.apify.com/api/v2#/reference/request-queues)
 
+## Status messages
+
+[Status messages](/platform/actors/development/programming-interface/status-messages) are lightweight, human-readable progress indicators displayed with the Actor run in the Apify Console (separate from logs). Use them to communicate high-level phases or milestones, such as "Fetching list", "Processed 120/500 pages", or "Uploading results".
+
+Update the status only when the user's understanding of progress changes - avoid frequent updates for every processed item. Detailed information should go to logs or storages (Dataset, Key-value store) instead.
```
barjin, line 206: nit: the example code is showing exactly this :)

docs/02_concepts/01_actor_lifecycle.mdx
```diff
@@ -242,6 +199,18 @@ When the <ApiLink to="apify/class/Dataset">`Dataset`</ApiLink> is stored on the
...
+
+Use <ApiLink to="class/Actor#setStatusMessage">`Actor.setStatusMessage()`</ApiLink> to set the message:
+
+<CodeBlock language="js">{StatusMessageSource}</CodeBlock>
+
+The SDK only sends an API request when the message text changes, so repeating the same message incurs no additional cost.
```
barjin, line 212: > The SDK only sends an API request when the message text changes

I don't think this is true. We need to send the request every time, as somebody else might have updated the message in the meantime.

docs/02_concepts/01_actor_lifecycle.mdx
```diff
@@ -242,6 +199,18 @@ When the <ApiLink to="apify/class/Dataset">`Dataset`</ApiLink> is stored on the
 - [Datasets API reference](https://docs.apify.com/api/v2#/reference/datasets)
 - [Request queues API reference](https://docs.apify.com/api/v2#/reference/request-queues)
 
+## Status messages
+
+[Status messages](/platform/actors/development/programming-interface/status-messages) are lightweight, human-readable progress indicators displayed with the Actor run in the Apify Console (separate from logs). Use them to communicate high-level phases or milestones, such as "Fetching list", "Processed 120/500 pages", or "Uploading results".
```
barjin, line 204: > (separate from logs)

technically half-true, setting the status message with SDK actually logs the status message.

https://github.com/apify/apify-sdk-js/blob/2c522d1d46f38d45fe8e3ff1fcb5d34b5dae1b6c/src/actor.ts#L1095-L1100

barjin (commented):
Thank you @TC-MO !

Here are my few first-minute ideas:

docs/02_concepts/01_actor_lifecycle.mdx
TC-MO: Is this valid also for Python SDK? I based my prose on Python SDK section about Status Messages so it might also be wrong for Python SDK?

docs/02_concepts/01_actor_lifecycle.mdx
TC-MO: Is this valid also for Python SDK? I based my prose on Python SDK section about Status Messages so it might also be wrong for Python SDK?

barjin: I believe so, the [code](https://github.com/apify/apify-sdk-python/blob/fbbf06eac1aacd8a9b880b4828fb8422413b251b/src/apify/_actor.py#L1296) doesn't seem to indicate anything like this.

I see this was added to the Python SDK docs by @vdusek - do you have any idea what this was referring to, Vlada?

docs/02_concepts/01_actor_lifecycle.mdx
barjin: It seems that Python logs the status message only when running locally, i.e., not on the platform. So it does make a little more sense for Python (it's not really that big of a deal for JS anyway).

B4nan (commented):
lets also make those changes in the last docs snapshot

https://github.com/apify/apify-sdk-js/tree/master/website/versioned_docs/version-3.4/02_concepts

docs/02_concepts/01_actor_lifecycle.mdx
vdusek: Yeah, I remember discussing this 1-2 years ago. And you are right. The status message should appear in the console with Actor runs, while locally it should be written to the logs (since there's no console to render it, hah).

docs/02_concepts/01_actor_lifecycle.mdx
vdusek: You're right, in Python SDK, the `set_status_message()` calls `self.apify_client.run(...).update(status_message=...)` every time and the sentence "The SDK only sends an API request when the message text changes" is wrong.

PR fixing it -> https://github.com/apify/apify-sdk-python/pull/859/

PR #588, opened 2026-04-15
docs: remove changelog from sidebar

TC-MO:


patrikbraborec (approved):

barjin (commented):
Note that these changes cause the sidebar not to render on the Changelog page, which imo looks a bit iffy (especially if you get there from any other page which has the sidebar layout).

Image

Image

I believe you can use the `displayed_sidebar` frontmatter field to enforce rendering the sidebar even if Changelog is not its item ([docs](https://docusaurus.io/docs/sidebar/multiple-sidebars#sidebar-association)).

TC-MO:
Yup, thank you I made the same mistake over all PR's so will be fixing it across the board to enforce sidebar on changelog

TC-MO:
Question: can we unify the .gitignore approach?

I applied the sidebar fix in CLI first (https://github.com/apify/apify-cli/pull/1083) - wrapping the sidebar in a named export so we can use `displayed_sidebar` frontmatter on changelog pages.

Applying the same fix here, I noticed this repo has multiple `.gitignore` files (`/docs/.gitignore`, root `.gitignore`) that blanket-ignore `changelog.md` at any depth. This makes the versioned changelogs untrackable, which blocks the `displayed_sidebar` fix.

Could we consolidate to a single root `.gitignore` entry targeting just `docs/changelog.md` (the generated one) instead? That way versioned changelogs stay tracked and we avoid redundant `.gitignore` files.

barjin:
If I recall correctly, we wanted to have the same changelog in all the docs versions (it's "versioned" by definition, and it might be confusing for the users to find an old version of the changelog with some versions missing). But if the docs team is fine with this (do we do this in CLI already?), it's fine, I guess. 

There's no deeper reason behind the multiple `.gitignore` files, I think.

TC-MO:
In CLI we had singular `.gitignore`(I guess a lot fo things are not standarized re: docs between the repos) with the fix to use `displayed_sidebar` I think by default people on older version would be served the snapshot `changelog.md` up until that version, not the main one, due to how docusaurus resolves versioned docs & sidebars (Unless I misunderstood that logic and their docs) Not sure if we can achieve always main sidebar without some custom work

wait, unless I'm overthinking it? and the actual way is to just add `displayed_sidebar` to the generated changelog instead of all the snapshots etc?

barjin:
> and the actual way is to just add displayed_sidebar to the generated changelog

That's what I would expect :+1: note that we likely don't want this directly in the changelog (might be weird for people viewing the changelog from outside of Docusaurus). 

Regarding the first part of your message - afaik we're currently not versioning changelog in SDK docs (https://docs.apify.com/sdk/js/docs/3.1/changelog is the same as https://docs.apify.com/sdk/js/docs/next/changelog)... and I thought we wanted that - but idk, you guys know more about DX :)

TC-MO:
Converting to draft for now

barjin (approved):
The changes look fine to me. Thank you @TC-MO !

PR #589, opened 2026-04-17
docs: Add the docs for missing versions 3.5, 3.6, and 3.7

nmanerikar:
As reported here https://apify.slack.com/archives/C0L33UM7Z/p1776335178659909

The SDK JS Overview page says the latest version is 3.4 while it is already 3.7

This PR adds the docs for the missing versions (and populates that dropdown)

I generated the docs for the following versions - 3.5.3 -> in the list as 3.5, 3.6, 3.7
I did this by checking out the respective tags and running each time:
```bash
$ npx docusaurus docs:version x.x
$ npx docusaurus api:version x.x
```

I hope this is the right way to do it!

Testing:
- from the `website` directory I ran `npm run build` and `npm run start`
- verified that the docs show up correctly 

Note: there were some build errors about broken links (not sure how to handle those, as the docs are auto-generated). But the docs loaded fine.


barjin (commented):
Thank you @nmanerikar !

> I did this by checking out the respective tags and running each time ...

Yup, I believe this is the way 👍 

The build seems to be failing because of absolute links in the markdown files (which resolve directly to HTTP pathnames, without adding the version). I suppose some of the paths changed in the most recent versions (so old versions linking to these documents will end up with broken links). I'll deal with it locally.

Thanks again!

B4nan:
Let's also add the CI step to generate this with feature releases automatically.

.github/workflows/release.yaml
```diff
@@ -140,16 +140,15 @@ jobs:
 
     version_docs:
         name: Version docs
-        if: ${{ github.event.inputs.release_type == 'minor' || github.event.inputs.release_type == 'major' }}
+        if: ${{ inputs.release_type == 'minor' || inputs.release_type == 'major' }}
         needs: [update_changelog]
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-22.04
```
B4nan, line 145: lets use 24 right ahead if we are pinning it

barjin: The rest of the jobs in the file is using `22.04` as of now, I'd rather bump them all together separately (for consistency, e.g., caching `node_modules`).

B4nan: fair enough, we should bump it everywhere

B4nan: lol, i thought this is a node version, not ubuntu version, i can't read anymore 😭

barjin: So... do you agree with bumping it in a separate PR? :smile:  If I recall correctly, `ubuntu-24.04` is having issues with our Playwright tests in the CI because of some AppArmor config... but we're not really running Playwright in SDK tests iirc

B4nan: I am fine leaving this for later, especially if there are issues with pw. I want to have node 24 everywhere, but I guess we are there already.

PR #592, opened 2026-04-24
feat: add subdivisionCode to proxy configuration

daniil-poletaev:
This PR adds new variable `subdivisionCode` and `apifyProxySubdivision` to Apify proxy configuration. This variable is currently needed for US state targeting.

More context: https://apify.slack.com/archives/C010Q0FBYG3/p1776704073278519

Same PR in python: https://github.com/apify/apify-sdk-python/pull/878

### DIFF
--- a/src/proxy_configuration.ts
+++ b/src/proxy_configuration.ts
@@ -17,6 +17,7 @@ const MAX_SESSION_ID_LENGTH = 50;
 const CHECK_ACCESS_REQUEST_TIMEOUT_MILLIS = 4_000;
 const CHECK_ACCESS_MAX_ATTEMPTS = 2;
 const COUNTRY_CODE_REGEX = /^[A-Z]{2}$/;
+const SUBDIVISION_CODE_REGEX = /^[A-Z0-9]{2}$/;
 
 export interface ProxyConfigurationOptions extends CoreProxyConfigurationOptions {
     /**
@@ -217,8 +218,12 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
                 countryCode: ow.optional.string.matches(COUNTRY_CODE_REGEX),
                 apifyProxyCountry:
                     ow.optional.string.matches(COUNTRY_CODE_REGEX),
-                subdivisionCode: ow.optional.string,
-                apifyProxySubdivision: ow.optional.string,
+                subdivisionCode: ow.optional.string.matches(
+                    SUBDIVISION_CODE_REGEX,
+                ),
+                apifyProxySubdivision: ow.optional.string.matches(
+                    SUBDIVISION_CODE_REGEX,
+                ),
                 password: ow.optional.string,
                 tieredProxyUrls: ow.optional.array.ofType(
                     ow.array.ofType(ow.string),
@@ -463,7 +468,7 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
             parts.push(`session-${sessionId}`);
         }
         if (subdivisionCode) {
-            parts.push(`country-${countryCode}-${subdivisionCode}`);
+            parts.push(`country-${countryCode}_${subdivisionCode}`);
         } else if (countryCode) {
             parts.push(`country-${countryCode}`);
         }
--- a/src/proxy_configuration.ts
+++ b/src/proxy_configuration.ts
@@ -17,6 +17,7 @@ const MAX_SESSION_ID_LENGTH = 50;
 const CHECK_ACCESS_REQUEST_TIMEOUT_MILLIS = 4_000;
 const CHECK_ACCESS_MAX_ATTEMPTS = 2;
 const COUNTRY_CODE_REGEX = /^[A-Z]{2}$/;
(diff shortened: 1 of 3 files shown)


src/proxy_configuration.ts
```diff
@@ -17,6 +17,7 @@ const MAX_SESSION_ID_LENGTH = 50;
 const CHECK_ACCESS_REQUEST_TIMEOUT_MILLIS = 4_000;
 const CHECK_ACCESS_MAX_ATTEMPTS = 2;
 const COUNTRY_CODE_REGEX = /^[A-Z]{2}$/;
+const SUBDIVISION_CODE_REGEX = /^[A-Z0-9]{2}$/;
```
barjin: Note that, e.g., [United Kingdom](https://en.wikipedia.org/wiki/ISO_3166-2:GB) or [Belgium](https://en.wikipedia.org/wiki/ISO_3166-2:BE) do three-lettered codes

src/proxy_configuration.ts
```diff
@@ -226,13 +255,20 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
 
         const groupsToUse = groups.length ? groups : apifyProxyGroups;
         const countryCodeToUse = countryCode || apifyProxyCountry;
+        const subdivisionCodeToUse = subdivisionCode || apifyProxySubdivision;
         const hostname = config.get('proxyHostname');
         const port = config.get('proxyPort');
 
+        if (subdivisionCodeToUse && !countryCodeToUse) {
+            throw new Error(
+                '"subdivisionCode" requires "countryCode" to be set.',
```
barjin: Perhaps we should make this even more explicit (e.g., mention ProxyConfiguration in the error message), so the user knows immediately where this comes from

barjin (commented):
Thank you @daniil-poletaev ! 

Regarding the new separate option, I have another question. Are we planning to support passing multiple countries at some point? (e.g., "I want to scrape xyz.com either through Belgium or the Netherlands to make the proxy pool larger").

In such a case, separate country and subdivision code options would require us to change the API again, as we couldn't match the country and the subdivision code

```typescript
countryCode: ['BE', 'NL'],
subdivisionCode: ['BRU', 'AMS'] // which country do these match?
```

Compare to the alternative `countryCode: ["BE-BRU", "NL-AMS"]`, which disambiguates by default. 

But, if we're positive that we don't want to go this way, it's a rather theoretical problem :)

src/proxy_configuration.ts
daniil-poletaev: Good catch 👍 

Fixed it

daniil-poletaev:
> Regarding the new separate option, I have another question. Are we planning to support passing multiple countries at some point? (e.g., "I want to scrape xyz.com either through Belgium or the Netherlands to make the proxy pool larger").

Maybe, but tbh it still would be a breaking change and we can change format later if it would be needed, so I would keep the format as-is for now 🤔

src/proxy_configuration.ts
```diff
@@ -56,6 +57,19 @@ export interface ProxyConfigurationOptions extends CoreProxyConfigurationOptions
...
+    /**
+     * If set, all proxied requests will use IP addresses geolocated to the specified subdivision (e.g. US state).
+     * Requires `countryCode` to be set. The value must follow the ISO 3166-2 subdivision code format,
+     * e.g. `'CA'` for California when `countryCode` is `'US'`.
+     */
+    subdivisionCode?: string;
+
+    /**
+     * Same option as `subdivisionCode` which can be used to
+     * configurate the proxy by UI input schema. You should use the `subdivisionCode` option in your crawler code.
```
barjin:
```suggestion
     * Same option as `subdivisionCode` which can be used to
     * configure the proxy by UI input schema. You should use the `subdivisionCode` option in your crawler code.
```

barjin (approved):
Alright, I don't have any further objections then :) (see small typo below)

Thank you @daniil-poletaev !

barjin (commented):
Sorry for the last-minute note - can you also please make changes to https://docs.apify.com/sdk/js/docs/concepts/proxy-management , just like with the Python PR? Thanks!

barjin (approved):
no further comments, thank you @daniil-poletaev !

src/proxy_configuration.ts
```diff
@@ -17,6 +17,7 @@ const MAX_SESSION_ID_LENGTH = 50;
 const CHECK_ACCESS_REQUEST_TIMEOUT_MILLIS = 4_000;
 const CHECK_ACCESS_MAX_ATTEMPTS = 2;
 const COUNTRY_CODE_REGEX = /^[A-Z]{2}$/;
+const SUBDIVISION_CODE_REGEX = /^[A-Z0-9]{1,3}$/;
```
bliuchak, line 21: I would highly recommend add there a comment that explains that 1-3 range for an ISO standard that expects only 2 chars 😄 

This was something that confused me.

bliuchak (approved):
Nice

Issue #433, opened 2025-08-21
Deprecate KeyValueStore.getPublicUrl(key) and introduce KeyValueStore.getRecordPublicUrl(key)


danpoletaev:
### Feature

1. Deprecate the synchronous method `KeyValueStore.getPublicUrl(key: string)`.
2. Introduce a new asynchronous method `KeyValueStore.getRecordPublicUrl(key: string)`.

### Motivation

The current `KeyValueStore.getPublicUrl(key`) method has several issues that impact developer experience, performance, and maintainability 👇 
- Synchronous: The method is synchronous.
- Confusing name 🤔: With the recent addition of a public URL for the list of keys, the name `getPublicUrl` is ambiguous. 
- Lack of code reuse: The method manually constructs the URL string instead of calling the apify-client. This creates redundant logic.

Introducing an async method with a more specific name (`getRecordPublicUrl`) that leverages the apify-client will solve these issues.

### Ideal solution or implementation, and any additional constraints

1. Mark the existing `KeyValueStore.getPublicUrl(key)` method as deprecated. The deprecation message should advise users to switch to the new `getRecordPublicUrl` method.

2. Implement the new method async `getRecordPublicUrl(key: string)` within the `KeyValueStore` class. This new method will call the appropriate apify-client function, `keyValueStore(id).getRecordUrl(key)`, and return the resulting URL.

PR #595, opened 2026-04-30
fix: adapt SDK storage layer to crawlee v4 StorageClient interface

B4nan:
## Summary

Crawlee v4 reshaped the `StorageClient` interface, removed the cached `storageObject` from `KeyValueStore`, and made `getPublicUrl` async. The SDK still targeted the v3 shape and no longer compiled. This PR adapts:

- **New `ApifyStorageClient` adapter** wrapping `apify-client`'s legacy `dataset()/keyValueStore()/requestQueue()` accessors and exposing the factory methods (`createDatasetClient` / `createKeyValueStoreClient` / `createRequestQueueClient`) crawlee now requires. Name → ID resolution goes through each collection's `getOrCreate(name)`.
- `Actor.init` and `_openStorage` now wrap `this.apifyClient` in the adapter before handing it to crawlee/`StorageManager`.
- `KeyValueStore.getPublicUrl` is now `async`; the per-store `urlSigningSecretKey` is fetched on demand via the (private) `client.getMetadata()` rather than the removed `storageObject` cache. **URL-signing behaviour in platform mode is preserved.**
- `Actor.openRequestQueue` reads `totalRequestCount` via `client.getMetadata()` (the old `client.get()` is gone).
- `StorageManager.openStorage` signature is now `(class, id?, client?)` — dropped the trailing `this.config` argument.

## Known gaps tracked in this PR

`apify-client`'s `DatasetClient` / `KeyValueStoreClient` / `RequestQueueClient` don't yet implement v4-added members like `getMetadata` and `getRecordPublicUrl`. The adapter casts through with `as unknown as ...` for now — the proper fix is to bring `apify-client`'s resource client shapes into structural alignment with `@crawlee/types`, which is out of scope here.

The `KeyValueStoreInfo` type in `@crawlee/types` also doesn't declare `urlSigningSecretKey`, so the SDK uses a local intersection type to access it. Worth surfacing upstream eventually.

## Stacking

Depends on #583 (config redesign). Rebases cleanly onto v4 once that lands.

### DIFF
--- a/packages/apify/src/apify_storage_client.ts
+++ b/packages/apify/src/apify_storage_client.ts
@@ -1,0 +1,104 @@
+import type {
+    CreateDatasetClientOptions,
+    CreateKeyValueStoreClientOptions,
+    CreateRequestQueueClientOptions,
+    DatasetClient,
+    KeyValueStoreClient,
+    RequestQueueClient,
+    StorageClient,
+} from '@crawlee/types';
+import type { ApifyClient } from 'apify-client';
+
+import type { Configuration } from './configuration.js';
+
+type StorageType = 'Dataset' | 'KeyValueStore' | 'RequestQueue';
+
+const DEFAULT_ID_CONFIG_KEY = {
+    Dataset: 'defaultDatasetId',
+    KeyValueStore: 'defaultKeyValueStoreId',
+    RequestQueue: 'defaultRequestQueueId',
+} as const;
+
+/**
+ * Bridges `apify-client`'s synchronous resource accessors (`dataset(id)`,
+ * `keyValueStore(id)`, `requestQueue(id, options?)`) to crawlee v4's
+ * `StorageClient` interface (async factory methods accepting either an `id`
+ * or a `name`).
+ *
+ * `storageExists()` is implemented so that `Dataset.open(idOrName)` and friends
+ * resolve a string argument to an id first (when one with that id exists on
+ * the platform) and fall back to a name otherwise — without this, crawlee's
+ * `resolveStorageIdentifier` would treat every string as a name and the SDK
+ * would silently create a brand-new storage whose name equals the passed-in id.
+ *
+ * When only a `name` is provided to a `create*Client` method, it is resolved
+ * to a concrete id via `getOrCreate(name)` — same behaviour the SDK relied on
+ * in v3.
+ */
+export class ApifyStorageClient implements StorageClient {
+    constructor(
+        private readonly client: ApifyClient,
+        private readonly config?: Configuration,
+    ) {}
+
+    async storageExists(id: string, type: StorageType): Promise<boolean> {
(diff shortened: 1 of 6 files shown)


packages/apify/src/apify_storage_client.ts
```diff
@@ -0,0 +1,68 @@
...
+    constructor(private readonly client: ApifyClient) {}
+
+    async createDatasetClient(
+        options?: CreateDatasetClientOptions,
+    ): Promise<DatasetClient> {
+        const id =
+            options?.id ??
+            (options?.name
+                ? (await this.client.datasets().getOrCreate(options.name)).id
+                : undefined);
```
barjin: Note that if we implement [`storageExists`](https://github.com/apify/crawlee/blob/712a5d1f20e67f3ff9d4ffee6da073045418770d/packages/types/src/storages.ts#L382), Crawlee should be able to resolve the identifiers on its own(?)

packages/apify/src/key_value_store.ts
```diff
@@ -15,24 +21,35 @@ export class KeyValueStore extends CoreKeyValueStore {
...
      * Returns a URL for the given key that may be used to publicly
      * access the value in the remote key-value store.
+     *
+     * On the Apify platform the URL is signed with the store's
+     * `urlSigningSecretKey` so that anyone with the URL can read the record
+     * without authentication. Locally we delegate to crawlee's default
+     * implementation (which produces a `file://` URL or returns `undefined`).
      */
-    override getPublicUrl(key: string): string {
+    override async getPublicUrl(key: string): Promise<string | undefined> {
```
barjin, line 30: Maybe this change closes [this issue](https://github.com/apify/apify-sdk-js/issues/433) - I'm not sure if the renaming was a requirement, or a way to make this BC.

packages/apify/src/key_value_store.ts
```diff
@@ -15,24 +21,35 @@ export class KeyValueStore extends CoreKeyValueStore {
...
         }
 
         const publicUrl = new URL(
             `${config.apiPublicBaseUrl}/v2/key-value-stores/${this.id}/records/${key}`,
         );
 
-        if (this.storageObject?.urlSigningSecretKey) {
+        // `client` is `private` on `CoreKeyValueStore`; bypass the visibility
+        // check to fetch the per-store secret. There is no public crawlee API
+        // surface for this yet — track upstream exposure as a follow-up.
```
barjin: Good point, please create the issue in Crawlee (I see that `Dataset` and `RequestProvider` already both have `client` public, so it's weirdly asymmetrical now).

barjin (approved):
A few ideas, but lgtm overall. Thanks!

@janbuchar might have more ideas as the original author

janbuchar:
@B4nan did you check parity with apify-sdk-python? I assume that there will be gaps in the request queue implementation, which is fine at this point.

PR #596, opened 2026-04-30
fix: adapt SDK ProxyConfiguration to crawlee v4 API

B4nan:
## Summary

Crawlee v4 reshaped `ProxyConfiguration`:
- `newProxyInfo` / `newUrl` now take a single `TieredProxyOptions` argument; the `(sessionId, options)` pair is gone.
- The protected `_handleCustomUrl(sessionId)` helper was removed.
- `_callNewUrlFunction` / `_handleTieredUrl` take options only.
- `ProxyInfo` (in `@crawlee/types`) no longer carries `sessionId`.

This PR adapts the SDK's override:
- `newProxyInfo` and `newUrl` accept `string | number | TieredProxyOptions | undefined` — existing SDK callers that pass a raw `sessionId` keep working, and the override is also compatible with crawlee's v4 single-options signature. A small `parseSessionIdOrOptions` helper discriminates and pulls `sessionId` from `options.request` when no explicit one is given.
- Inlined custom-URL session stickiness as `getSessionIndex(sessionId)` (replacing the removed `_handleCustomUrl`), keyed on the inherited `usedProxyUrls` map.
- Re-declared `sessionId?: string` on the SDK's `ProxyInfo` interface so users can keep reading `proxyInfo.sessionId`.
- `ProxyInfo` is now imported from `@crawlee/types` (no longer re-exported from `@crawlee/core`).
- Tightened a `.some(url => url.includes(...))` for the new `(string | null)[]` shape.

## Stacking

Depends on #583 (config redesign). Rebases cleanly onto v4 once that lands.

### DIFF
--- a/packages/apify/src/proxy_configuration.ts
+++ b/packages/apify/src/proxy_configuration.ts
@@ -1,8 +1,6 @@
-import type {
-    ProxyConfigurationOptions as CoreProxyConfigurationOptions,
-    ProxyInfo as CoreProxyInfo,
-} from '@crawlee/core';
+import type { ProxyConfigurationOptions as CoreProxyConfigurationOptions } from '@crawlee/core';
 import { ProxyConfiguration as CoreProxyConfiguration } from '@crawlee/core';
+import type { ProxyInfo as CoreProxyInfo } from '@crawlee/types';
 import { gotScraping } from 'got-scraping';
 import ow from 'ow';
 
@@ -12,11 +10,16 @@ import { cryptoRandomObjectId } from '@apify/utilities';
 import { Actor } from './actor.js';
 import { Configuration } from './configuration.js';
 
-// https://docs.apify.com/proxy/datacenter-proxy#username-parameters
-const MAX_SESSION_ID_LENGTH = 50;
 const CHECK_ACCESS_REQUEST_TIMEOUT_MILLIS = 4_000;
 const CHECK_ACCESS_MAX_ATTEMPTS = 2;
 const COUNTRY_CODE_REGEX = /^[A-Z]{2}$/;
+
+// Apify Proxy session identifier embedded in the proxy username — opaque to
+// users; a fresh one is minted for every URL the SDK hands out so that the
+// returned proxy URLs are independent.
+const SESSION_ID_LENGTH = 12;
+
+type NewUrlOptions = Parameters<CoreProxyConfiguration['newProxyInfo']>[0];
 
 export interface ProxyConfigurationOptions
     extends CoreProxyConfigurationOptions {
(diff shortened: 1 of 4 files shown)


test/apify/proxy_configuration.test.ts
```diff
@@ -714,70 +671,7 @@ describe('Actor.createProxyConfiguration()', () => {
...
-                [
-                    'http://groups-GROUP3+GROUP4,country-DE:password@proxy.apify.com:8000',
-                ],
-                ['http://groups-GROUP3+GROUP4:password@proxy.apify.com:8000'],
-            ]);
-        });
-    });
+    // `tieredProxyUrls` / `tieredProxyConfig` were removed from
+    // crawlee v4 (apify/crawlee#3599); the corresponding test groups
+    // were dropped here too.
```
barjin: nit: can we do away with these "gravestone" comments? 

These are useful when reading the AI output, but I don't think we should commit these. Future maintainers won't care about the tests that are not here anymore.

packages/apify/src/proxy_configuration.ts
```diff
@@ -304,36 +320,64 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
      * @return Represents information about used proxy and its configuration.
      */
     override async newProxyInfo(
-        sessionId?: string | number,
-        options?: Parameters<CoreProxyConfiguration['newProxyInfo']>[1],
+        sessionIdOrOptions?:
+            | string
+            | number
+            | Parameters<CoreProxyConfiguration['newProxyInfo']>[0],
+        legacyOptions?: Parameters<CoreProxyConfiguration['newProxyInfo']>[0],
```
barjin: The removal of the `sessionId` parameter from Crawlee v4 was intentional, see [this comment](https://github.com/apify/crawlee/pull/3199#discussion_r2452694371) .

According to the new "`UserPool`" design, `ProxyConfiguration` shouldn't care for `Session` details (the resolved proxy URL is stored in a `Session` _after_ it's retrieved and the `ProxyConfiguration` is not queried again). 

Imo, each call to `newProxyInfo` in SDK should just return a random, valid proxy URL - that is, e.g., with random session IDs. This way, the SDK shields the user from the Apify Proxy session implementation.

```typescript
proxyConfig.newProxyInfo() // { url: "http://session-131231@proxy.apify.com" }
proxyConfig.newProxyInfo() // { url: "http://session-234244@proxy.apify.com" }
proxyConfig.newProxyInfo() // { url: "http://session-342434@proxy.apify.com" }
// ...
```

As far as the consumer is concerned, these are just opaque URLs.

packages/apify/src/proxy_configuration.ts
```diff
@@ -362,40 +412,41 @@ export class ProxyConfiguration extends CoreProxyConfiguration {
...
+        if (!this.usedProxyUrls.has(sessionId)) {
+            this.usedProxyUrls.set(
+                sessionId,
+                this.proxyUrls![
+                    this.usedProxyUrls.size % this.proxyUrls!.length
+                ],
+            );
+        }
+        return this.proxyUrls!.indexOf(this.usedProxyUrls.get(sessionId)!);
     }
```
barjin: What is the reason behind this? Perhaps we can remove this, given the returned urls should be independent?

barjin (commented):
Thank you @B4nan ! Here, I have more talking points. 

Please see the comments below ⬇️

PR #602, opened 2026-05-05
docs: update codeowners

TC-MO:


TC-MO:
btw should it be possible to merge without approval here?

barjin (approved):
> btw should it be possible to merge without approval here?

Idk, I'd rather err on the side of caution in these projects (Client more so, as it's a cross-team effort). Here's your ✔️  though :)

PR #609, opened 2026-05-14
chore: Use custom action to commit changes in CI instead of `git commit`

fnesveda:
We want to enforce commit signing for all commits in our repositories.
To do that, we need to make sure even commits created by CI workflows are signed.

It would be possible to sign using GPG keys, but that would require a lot of maintenance.

Instead, we can commit using the GitHub GraphQL API, which automatically signs commits.

This PR replaces direct `git commit` / `git push` usage (and third-party commit actions like `EndBug/add-and-commit`)
with the `apify/actions/signed-commit` action, which uses the GraphQL API under the hood.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -40,21 +40,11 @@ jobs:
                   SMARTLOOK_PROJECT_KEY: ${{ secrets.SMARTLOOK_DOCS_PROJECT_KEY }}
                   SEGMENT_TOKEN: ${{ secrets.SEGMENT_TOKEN }}
 
-            - name: Stage the updated package.json and lockfile
-              id: stage
-              run: |
-                  git add website/package.json pnpm-lock.yaml
-                  if git diff --cached --quiet; then
-                    echo "has-changes=false" >> "$GITHUB_OUTPUT"
-                  else
-                    echo "has-changes=true" >> "$GITHUB_OUTPUT"
-                  fi
-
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yaml
```diff
@@ -41,13 +41,11 @@ jobs:
                   SEGMENT_TOKEN: ${{ secrets.SEGMENT_TOKEN }}
 
             - name: Commit the updated package.json and lockfile
-              uses: stefanzweifel/git-auto-commit-action@v7
```
janbuchar, line 44: Huh, an oddball 🙂

janbuchar (approved):

B4nan (approved):

.github/workflows/release.yaml
```diff
@@ -81,7 +81,7 @@ jobs:
         needs: [release_metadata, build_and_test]
         runs-on: ubuntu-22.04
         outputs:
-            changelog_commitish: ${{ steps.commit.outputs.commit_long_sha || github.sha }}
+            changelog_commitish: ${{ steps.commit.outputs.commit_long_sha }}
```
janbuchar, line 84: Will this be an issue in any of the other repositories? 🤔

fnesveda: I fixed it everywhere already 🙂

janbuchar: Very cool, thanks!

PR #636, opened 2026-06-04
refactor!: drop got-scraping, fs-extra, and ow dependencies

B4nan:
Drops three runtime dependencies in favor of native Node.js APIs and packages the SDK already pulls in. Part of the v4 dependency cleanup.

## Changes

| Dependency | Replaced with | Where |
|---|---|---|
| `got-scraping` | native `fetch` + `undici`'s `ProxyAgent` | Apify Proxy status check (`ProxyConfiguration._fetchStatus`) |
| `fs-extra` | `node:fs` (`JSON.parse(readFileSync(...))`) | `utils.ts` |
| `ow` | `zod` (already a dependency) | `Actor.addWebhook` / `setStatusMessage` / `openDataset` / `openKeyValueStore` / `openRequestQueue`, `ProxyConfiguration` |

`undici` is added as the one new dependency — it's needed for `ProxyAgent` (Node exposes native `fetch` but not a proxy dispatcher), is maintained by the Node core team, and is the same engine behind native `fetch`. Net direct-dependency count still goes **down**.

Also bumps `apify-client` to `^2.23.4`, which types `WebhookUpdateData.eventTypes` as a `readonly` array ([apify-client-js#931](https://github.com/apify/apify-client-js/pull/931)) — so `Actor.addWebhook` no longer copies the array.

## ⚠️ Breaking change — validation error messages

Argument validation switched from `ow` to `zod`, so the **error messages thrown for invalid arguments changed**. We now throw a plain, human-readable sentence (zod's `prettifyError`) instead of a JSON dump, and attach the structured `ZodError` as the error's `cause`:

```diff
- Expected property string `countryCode` to match `/^[A-Z]{2}$/`, got `CZE` in object   // v3 (ow)
+ ✖ Invalid string: must match pattern /^[A-Z]{2}$/                                       // v4 (zod)
+   → at countryCode
```

Affects `Actor.addWebhook()`, `Actor.setStatusMessage()`, `Actor.openDataset()` / `openKeyValueStore()` / `openRequestQueue()`, and the `ProxyConfiguration` constructor. Documented in the v4 upgrading guide; tests assert the exact new messages.

## Notes

- `ws` is intentionally **not** touched here — that refactor (→ native `WebSocket`) is platform-only testable and will be a separate follow-up.
- 122 unit tests pass; build + lint clean.

BREAKING CHANGE: argument validation switched from `ow` to `zod`, so the error messages thrown for invalid arguments to `Actor.addWebhook()`, `Actor.setStatusMessage()`, `Actor.openDataset()` / `openKeyValueStore()` / `openRequestQueue()`, and the `ProxyConfiguration` constructor have changed. The thrown `Error` now carries the structured `ZodError` as its `cause`. See the v4 upgrading guide.


package.json
```diff
@@ -80,12 +80,10 @@
...
         "@crawlee/types": "^4.0.0-beta.61",
         "@crawlee/utils": "^4.0.0-beta.61",
-        "apify-client": "^2.17.0",
-        "fs-extra": "^11.2.0",
-        "got-scraping": "^4.2.1",
-        "ow": "^2.0.0",
+        "apify-client": "^2.23.4",
         "semver": "^7.5.4",
         "tslib": "^2.6.2",
+        "undici": "^7.0.0",
```
janbuchar: Have you considered using `impit` instead? You'll most likely pull it with crawlee anyway.

janbuchar (approved):

package.json
B4nan: Yes, but it feels like an overkill, given the recent discussion about its use in the clients, I would rather not use it here either.

> You'll most likely pull it with crawlee anyway.

The end goal to me would be not depending on impit (or any other 3rd party dependency) because of crawlee. That's why I mentioned the possible `@crawlee/common` package in the other PR (which would be ideally dependency free).

janbuchar: Yeah, I realized the same a couple of minutes ago. Thanks.

PR #637, opened 2026-06-05
refactor: use native node:http for the proxy status check (drop undici)

B4nan:
Follow-up to #636. That PR replaced `got-scraping` with `undici`'s `ProxyAgent` for the Apify Proxy status check — but undici isn't actually needed.

The status endpoint (`http://proxy.apify.com/?format=json`) is plain HTTP fetched **through** an HTTP proxy, which native `node:http` handles directly: an absolute-form request URL plus a `Proxy-Authorization` header. No proxy-agent dependency.

It's written with modern primitives — no callback soup:

```ts
const request = httpRequest({ host: proxy.hostname, port: proxy.port, path: target.href, headers, signal: AbortSignal.timeout(...) });
request.end();
const [response] = await once(request, 'response'); // rejects on error/abort
// ...status-code check...
return await json(response); // node:stream/consumers
```

- `AbortSignal.timeout` for the timeout, `events.once()` for the response (rejects on connection error/abort → caught by the existing retry loop), `node:stream/consumers` `json()` to parse.
- All available on **Node 22.0**, so this reverts the `engines.node` bump back to `>=22.0.0` and **removes `undici`** — net result: dropping `got-scraping` cost us *zero* new dependencies.

The status request is encapsulated in `_requestStatus(statusUrl, proxyUrl)`; tests spy on that seam (and assert the proxy URL it's routed through), so no real network calls.

Caveat: this cleanly covers the default HTTP status URL. If someone overrides `APIFY_PROXY_STATUS_URL` to an `https://` URL (non-default, unusual), the forward-proxy GET won't tunnel it and the check falls back to its existing non-fatal "access check timed out" warning. Happy to add CONNECT tunneling if we think that's worth supporting.

Compile, lint, build, and 123 unit tests pass.

### DIFF
--- a/docs/04_upgrading/upgrading_v4.md
+++ b/docs/04_upgrading/upgrading_v4.md
@@ -5,9 +5,9 @@ title: Upgrading to v4
 
 This page summarizes the breaking changes between Apify SDK v3 and v4. Apify SDK v4 adopts the redesigned Crawlee v4 interfaces (`Configuration`, `EventManager`, `StorageClient`, `ProxyConfiguration`), so most of the changes here track the corresponding Crawlee v4 changes.
 
-## Node.js 22.19+
+## Node.js 22+
 
-The SDK now requires **Node.js 22.19 or newer**.
+The SDK now requires **Node.js 22 or newer**.
 
 ## ESM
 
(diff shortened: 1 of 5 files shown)


B4nan:
@janbuchar @barjin I've opened this as a possible follow up, not sure if we want/care about depending on undici here, given we need it for this single check, maybe better to go native?

barjin (approved):
Thanks @B4nan !

I looked a bit deeper into the whole proxy check.

It seems that the check should just send a(n authorized) `GET http://proxy.apify.com/?format=json HTTP/1.1` request 

| Image |
|---|
| `got-scraping` on `v3` |

See that the current `ProxyAgent` approach `CONNECT`s _through_ the proxy back to `?format=json` - but the second, inner request is not authorized (and accesses the wrong path), so it fails.

| Image |
|---|
| `fetch` with `ProxyAgent`  on `v4`|

This PR fixes it again ⬇️ 

| Image |
|---|
| `refactor/proxy-status-native-http` |

The concern with HTTPS proxies and `CONNECT` method from the PR description is, imo, not too pressing. 

On the other hand, we should check this on the platform (the proxy URL is, afaik, different from inside the Actors to speed up the networking), but I don't see why it shouldn't work now 👍

PR #542, opened 2026-01-16
refactor: reorganize documentation structure

patrikbraborec:
## Summary
- Restructured documentation from flat structure to numbered hierarchical organization (01_introduction, 02_concepts, 03_guides, 04_upgrading)
- Removed old flat structure files and updated sidebars configuration
- Improved navigation and content organization while maintaining all existing content

## Test plan
- [ ] Verify all documentation pages render correctly with new structure
- [ ] Check that navigation in sidebars works as expected
- [ ] Confirm no broken links or missing content
- [ ] Test documentation build process completes successfully

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #638, opened 2026-06-09
docs: add docs link checker to PRs, fix broken links

barjin:
Follow-up after https://github.com/apify/apify-client-js/pull/858

### DIFF
--- a/.github/workflows/test-and-release.yaml
+++ b/.github/workflows/test-and-release.yaml
@@ -80,6 +80,57 @@ jobs:
               run: pnpm build
               working-directory: ./website
 
+            - name: Run Lychee Link Checker
+              id: lychee
+              if: github.event_name == 'pull_request'
+              uses: lycheeverse/lychee-action@v2.8.0
+              with:
+                  fail: false
+                  args: >
+                      --base-url https://docs.apify.com
+                      --max-retries 6
+                      --exclude="github.com"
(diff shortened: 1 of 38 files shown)


barjin:
Screenshot from 2026-06-10 13-50-44


@patrikbraborec It seems the `type-script-actor` => `typescript-setup` swap from https://github.com/apify/apify-sdk-js/pull/542 is unfinished. The document's `frontmatter` still contains `id: type-script-actor`, which emits the `type-script-actor.html` file in the build (see [docs on ids and slugs](https://docusaurus.io/docs/create-doc#document-id)).

At the same time, the `nginx` setup redirects `...../type-script-actor` to `...../typescript-setup`, which results in 404. You can check this by clicking through the links in the SDK docs to the Typescript actor page (will work), then reloading the page with Ctrl+R, which will show 404.

What is the intended behaviour? `..../typescript-setup` being the only valid path? If so, can we update the id in this PR?

barjin:
The last commits implement the fix from https://github.com/apify/apify-sdk-js/pull/638#issuecomment-4669911776

The broken links to the Client docs come from the re-exported Crawlee symbols. This has been fixed [here](https://github.com/apify/crawlee/pull/3730) (will come into force on the next stable Crawlee release).

janbuchar (approved):

patrikbraborec (approved):

PR #643, opened 2026-06-16
feat: implement crawlee v4 RequestQueueClient over apify-client

B4nan:
Crawlee v4 redesigned the request queue storage layer into a stateful, pull-based `RequestQueueBackend` interface, modeled on the Python SDK. The SDK's old name-remapping proxy can't satisfy it, so this PR implements it for the Apify platform, with two access modes. The Python SDK's request queue clients were the starting point, but the design follows the JS API: the crawlee JS frontend already keeps large dedup caches, so the backends only manage what the frontend can't.

## Request queue access modes

`single` (default, `ApifyRequestQueueSingleBackend`) assumes the run is the only consumer of the queue. There is no request locking; the queue head is estimated locally, and requests added by this client are served from a local cache, so a typical request costs no read API calls. On the first add, up to 10k existing requests are prefetched, letting resurrected runs deduplicate re-added requests locally instead of paying a platform write per request. Multiple producers may still add requests concurrently.

`shared` (`ApifyRequestQueueSharedBackend`) is safe for any number of concurrent consumers. Fetched requests are locked server-side (`listAndLockHead`), and the lock duration follows the crawler's expected request processing time (`setExpectedRequestProcessingTimeSecs`, i.e. handler timeout plus padding). Reclaims release the lock so other consumers can pick the request up immediately, and `isFinished` consults `queueHasLockedRequests` so no consumer exits while another still holds work.

The mode is selected per storage backend, as in the Python SDK: `Actor.init({ requestQueueAccess: 'shared' })` or `new ApifyStorageBackend(client, { requestQueueAccess: 'shared' })`. Opening the same queue in both modes at once is not supported; whichever backend opens it first wins (documented on `getStorageBackendCacheKey`).

Request queue clients now send a stable per-run `clientKey` (the run id, same as the Python SDK), so `hadMultipleClients` stays meaningful and a migrated or resurrected run re-acquires the locks of its previous incarnation.

Both backends honor the beta.105 contract: `markRequestAsHandled`/`reclaimRequest` of a request that does not exist return `undefined` without upserting it, and `purge()` throws, since the platform has no truncate endpoint (Python behaves the same).

## Review feedback addressed

- `ApifyStorageClient` renamed to `ApifyStorageBackend` (class and file), `OpenStorageContext.client` renamed to `backend`, stray comment removed.
- The `single`/`shared` toggle from the review is implemented here rather than tracked in an issue.

## Crawlee beta.105 adaptation (was beta.71)

- `create*Backend` now receives crawlee's `StorageIdentifier` (`{id} | {name} | {alias}`). The backend resolves aliases itself: `__default__` maps to the run's default storage, other aliases resolve via `ACTOR_STORAGES_JSON`, undeclared aliases on the platform get a clear error, and anywhere else an unnamed storage is created per alias and process. A KVS-persisted alias mapping like Python's `AliasResolver`, which would let undeclared aliases survive migrations, is left as a follow-up.
- The key-value store backend is now a byte transport: `getValue` reads buffers and leaves parsing to crawlee's frontend (previously the value got parsed twice), `listKeys` returns the new paginated shape, and `recordExists` passes through.
- `getStorageBackendCacheKey` partitions crawlee's storage cache by API base URL and token.
- Mechanical drift: `getGlobalConfig` renamed to `getGlobalConfiguration`; `QueueOperationInfo`/`Constructor`/`Dictionary` moved to `@crawlee/types`; `snakeCaseToCamelCase` inlined (removed upstream); `ProxyConfiguration` internals were privatized upstream (own `log`, no more base-class field reads); `KeyValueStore.config`/`client` fields replaced by `serviceLocator`/`backend`.

## Validation

Unit tests cover both backends against a mocked `apify-client` (head estimation, dedup, prefetch, in-progress tracking, forefront ordering, lock duration raising, `queueHasLockedRequests`, the no-upsert contract), plus the storage backend's alias resolution and access mode wiring. Build, lint, and the full suite (148 tests) pass. The `single` mode is the successor of the flow validated end-to-end on the platform earlier in this PR; re-validating both modes on the platform against beta.105 is the remaining step.


src/storage.ts
```diff
@@ -137,7 +137,7 @@ function resolveStorageIdentifier(
 
 export interface OpenStorageContext {
     config: Configuration;
-    client?: StorageClient;
+    client?: StorageBackend;
```
janbuchar: I guess the field deserves a rename too?

src/apify_storage_client.ts
```diff
@@ -200,13 +201,15 @@ export class ApifyStorageClient implements StorageClient {
...
+        ) as unknown as KeyValueStoreBackend;
     }
 
-    async createRequestQueueClient(options?: CreateRequestQueueClientOptions): Promise<RequestQueueClient> {
+    async createRequestQueueBackend(options?: CreateRequestQueueBackendOptions): Promise<RequestQueueBackend> {
         const id = await this.resolveId(options, 'RequestQueue');
         const client = this.client.requestQueue(id, options?.clientKey ? { clientKey: options.clientKey } : undefined);
-        return adapt(client, { getMetadata: 'get', drop: 'delete' }, noPurge) as unknown as RequestQueueClient;
+        // Crawlee v4's RequestQueueBackend is a stateful, pull-based interface that
+        // can't be satisfied by name-remapping; it's implemented on apify-client.
```
janbuchar:
```suggestion
```

ok thx bye

src/apify_storage_client.ts
```diff
@@ -153,7 +154,7 @@ function adapt<T extends object>(
  * const { items } = await dataset.getData();
  * ```
  */
-export class ApifyStorageClient implements StorageClient {
+export class ApifyStorageClient implements StorageBackend {
```
janbuchar: Rename the class and file please

src/apify_storage_client.ts
```diff
@@ -200,13 +201,15 @@ export class ApifyStorageClient implements StorageClient {
                 // crawlee expects an array; apify-client returns `{ items }`.
                 listKeys: async (opts?: Parameters<typeof client.listKeys>[0]) => (await client.listKeys(opts)).items,
             },
-        ) as unknown as KeyValueStoreClient;
+        ) as unknown as KeyValueStoreBackend;
     }
 
-    async createRequestQueueClient(options?: CreateRequestQueueClientOptions): Promise<RequestQueueClient> {
+    async createRequestQueueBackend(options?: CreateRequestQueueBackendOptions): Promise<RequestQueueBackend> {
```
janbuchar: The options don't make it possible to toggle between `single` and `shared` modes (no locking vs locking) — is this tracked in an issue? Or shall we fix it right here right now?

janbuchar (changes requested):
The description seems outdated, e.g.:

> - **`ApifyStorageClient.setStatusMessage`** — Crawlee v4 calls this on the storage client to surface crawl progress (it was warning `setStatusMessage is not a function`).

is no longer true since https://github.com/apify/crawlee/pull/3818

src/apify_storage_backend.ts
```diff
@@ -179,34 +252,54 @@ export class ApifyStorageClient implements StorageClient {
...
             noPurge,
-        ) as unknown as DatasetClient;
+        ) as unknown as DatasetBackend;
     }
 
-    async createKeyValueStoreClient(options?: CreateKeyValueStoreClientOptions): Promise<KeyValueStoreClient> {
+    async createKeyValueStoreBackend(options?: StorageIdentifier): Promise<KeyValueStoreBackend> {
         const id = await this.resolveId(options, 'KeyValueStore');
         const client = this.client.keyValueStore(id);
         return adapt(
```
janbuchar: I don't know, it would be easier to understand if we just made two dummy classes, `DatasetBackend` and `KeyValueStoreBackend`, like for the request queue

janbuchar (approved):
No sense in blocking this, let's try it out in the wild

src/apify_storage_backend.ts
B4nan: Good point, done via d1b71f6d2a873f217e28119bdec788a35b830adc

src/apify_storage_backend.ts
```diff
@@ -0,0 +1,342 @@
...
+ * Bridges `apify-client`'s synchronous resource accessors (`dataset(id)`,
+ * `keyValueStore(id)`, `requestQueue(id, options?)`) to crawlee v4's
+ * `StorageBackend` interface (async factory methods accepting an `id`,
+ * a `name`, or an `alias`).
+ *
+ * For the run's default dataset it transparently swaps in a charging-aware
+ * dataset client (pay-per-event on `Actor.pushData()`), provided a charging
+ * manager is supplied and a default-dataset-item price is configured.
+ *
+ * `storageExists()` lets `Dataset.open(idOrName)` resolve a string to an id
```
janbuchar: Implementation details leaking in comment?

test/apify/apify_storage_backend.test.ts
```diff
@@ -0,0 +1,251 @@
...
+
+type MockApifyClient = ReturnType<typeof createMockApifyClient>;
+
+function asApifyClient(mock: MockApifyClient): ApifyClient {
+    return mock as unknown as ApifyClient;
+}
+
+describe('ApifyStorageBackend', () => {
+    test('creates a single-consumer request queue backend by default', async () => {
+        const client = createMockApifyClient();
```
janbuchar, line 40: This doesn't seem to test anything typescript wouldn't catch

janbuchar (approved):

PR #653, opened 2026-06-30
fix: allow Actor.newClient().dataset() without Actor.init()

B4nan:
## What & why

The charging-aware client returned by `Actor.newClient()` throws `Error: ChargingManager is not initialized` from its `dataset()` getter when `Actor.init()` hasn't been called:

```js
const client = Actor.newClient({ token });
client.dataset(someId); // 💥 ChargingManager is not initialized
```

`PatchedApifyClient.dataset()` always evaluated `actor.getChargingManager().getPricingInfo()`, and `getPricingInfo()` throws when the manager isn't initialized. But the [docs](https://docs.apify.com/sdk/js/reference/class/Actor#init) state `Actor.init()` is only required on the platform, so a standalone client should work. This regressed standalone usage (it surfaced in Crawlee's e2e harness, which reads dataset results via `Actor.newClient()` without `init()`).

## Fix

Mirror the guard the SDK already applies in the `pushData` charging path (`Actor.pushData()` works without `init()` — charging just isn't configured): only apply charging interception when the `ChargingManager` is initialized; otherwise return a plain `DatasetClient`.

- Add `ChargingManager.isInitialized` (mirrors the existing `chargingState === undefined` guards) and gate the interception on it.
- Hand the patched client the raw manager instead of the public `getChargingManager()`. The public getter also logs a misleading *"did you forget Actor.init()?"* warning on this supported path **and** latches `warnedAboutMissingInitCall`, which would make a subsequent `Actor.init()` throw.

When initialized (platform run, PPE pricing, default dataset) behavior is unchanged — the existing charging suite passes.

## Test

Added a regression test (`dataset()` works without calling `Actor.init()`) that asserts no throw and no spurious init warning.

### DIFF
--- a/src/actor.ts
+++ b/src/actor.ts
@@ -1497,7 +1497,12 @@ export class Actor<Data extends Dictionary = Dictionary> {
                 ...storageClientOptions,
                 ...options, // allow overriding the instance configuration
             },
-            this,
+            {
+                config: this.config,
+                // Hand over the raw manager rather than the public getChargingManager(), which
+                // warns about a missing init() — a standalone client is a supported use case.
+                getChargingManager: () => this.chargingManager,
+            },
         );
     }
 
--- a/src/charging.ts
+++ b/src/charging.ts
@@ -236,6 +236,14 @@ export class ChargingManager {
     }
 
     /**
+     * Whether {@link ChargingManager.init} has run. All charging operations (including
+     * {@link ChargingManager.getPricingInfo}) require an initialized manager.
+     */
+    get isInitialized(): boolean {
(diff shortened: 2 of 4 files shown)


janbuchar (approved):
If [E2E tests](https://github.com/apify/apify-sdk-js/actions/runs/28438160950) succeed, I think it's good to go. I do look forward to replacing this with something sane in v4.

barjin (approved):
lgtm, thanks @B4nan  👍

B4nan:
The tests are pretty flaky, rerun succeeded, so merging.

PR #667, opened 2026-07-16
fix: Fix `inherit` timeout units and no-timeout-left case

Pijukatel:
**Timeout units:**
Remaining time was calculated in milliseconds, while API accepts a timeout in seconds. This led to incorrect timeout values being passed to the API. Properly pass the timeout in seconds to the API.

**No timeout left case:**
Limit the minimal timeout to 1s, as the API interprets a timeout of 0 as no timeout at all.

### DIFF
--- a/src/actor.ts
+++ b/src/actor.ts
@@ -252,6 +252,8 @@ export interface Timeout {
      *
      * Using `inherit` will set timeout of the newly started Actor run to the time
      * remaining until this Actor run times out so that the new run does not outlive this one.
+     * If there is no time remaining (this run is already past its own timeout), the new run
+     * is not started at all — the method only logs a warning and returns `undefined`.
      */
     timeout?: number | 'inherit';
 }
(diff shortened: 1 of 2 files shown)


Pijukatel:
Parity with https://github.com/apify/apify-sdk-python/pull/1051

src/actor.ts
```diff
@@ -2299,13 +2299,17 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
-     * run was started without a timeout.
+     * Get time remaining from the Actor run timeout, rounded up to whole seconds with minimum value of 1 second.
+     *
+     * The API treats a 0 second timeout as no timeout, the minimum acceptable timeout is 1 second.
+     *
+     * Returns `undefined` if not on an Apify platform or the current run was started without a timeout.
      */
     private getRemainingTime(): number | undefined {
         const env = this.getEnv();
+        const smallestPossibleApiTimeout = 1;
```
barjin:
```suggestion
        const MINIMUM_API_TIMEOUT_SECS = 1;
```

src/actor.ts
```diff
@@ -2299,13 +2299,17 @@ export class Actor<Data extends Dictionary = Dictionary> {
...
     /**
-     * Get time remaining from the Actor run timeout. Returns `undefined` if not on an Apify platform or the current
-     * run was started without a timeout.
+     * Get time remaining from the Actor run timeout, rounded up to whole seconds with minimum value of 1 second.
+     *
+     * The API treats a 0 second timeout as no timeout, the minimum acceptable timeout is 1 second.
+     *
+     * Returns `undefined` if not on an Apify platform or the current run was started without a timeout.
      */
     private getRemainingTime(): number | undefined {
```
barjin:
```suggestion
     * Get time remaining from the Actor run timeout in seconds, rounded up to whole seconds with minimum value of 1 second.
     *
     * The API treats a 0 second timeout as no timeout, the minimum acceptable timeout is 1 second.
     *
     * Returns `undefined` if not on the Apify platform or the current run was started without a timeout.
     */
    private getRemainingTimeSecs(): number | undefined {
```

barjin (approved):
Thank you @Pijukatel ! 

Few naming nits to prevent similar confusion in the future, but nothing blocking 👍

PR #689, opened 2026-08-17
feat: helper turning `globs`/`pseudoUrls` inputs into a `transformRequestFunction`

### LINKED ISSUE #687: compatibility layer for v3 `enqueueLinks` options on Crawlee v4
### Goal
v3-shaped `globs` / `regexps` / `pseudoUrls` keep working on Crawlee v4 — including the per-pattern request options the Apify Console composes — without the actor changing any code.

### Approach

A small generic **hook in Crawlee**, filled by the **Apify SDK**. Crawlee never learns what a pseudo-URL is; all legacy knowledge lives in `apify` and can be removed later without a breaking change.

This supersedes the earlier shim proposal in apify/crawlee#3871, which put the whole mapping in Crawlee.

## Why a hook, and not the SDK alone

The SDK cannot intercept `enqueueLinks` from outside:

- ESM live bindings are read-only, so `@crawlee/core`'s exported `enqueueLinks` cannot be reassigned; the subclass crawlers import it directly.
- The context helper is built per-request inside `addHelpers()`, so there is no object to patch ahead of time.
- `extendContext` only reaches crawlers that delegate to the context's `enqueueLinks`. Cheerio (`cheerio-crawler.ts:262`) and browser (`browser-crawler.ts:578`) do; **jsdom does not** (`jsdom-crawler.ts:355`), and neither do `enqueueLinksByClickingElements` or `SitemapRequestLoader`. A layer that silently no-ops on `jsdom-scraper` is worse than none.

"Automatic" therefore requires something Crawlee owns to call the mapper. v4 already has the right seam: `serviceLocator`, the get/set registry for pluggable services, which the SDK already populates.

## Change 1 — Crawlee (~10 lines + 4 one-line call sites)

```ts
// service_locator.ts
getEnqueueLinksOptionsPreprocessor(): (o: EnqueueLinksOptions) => EnqueueLinksOptions;  // default: identity
setEnqueueLinksOptionsPreprocessor(fn): void;
```

Called as the first statement of each entry point, **before** `ow` validation:

```ts
options = serviceLocator.getEnqueueLinksOptionsPreprocessor()(options);
```

Four call sites — they validate independently, so each needs the line:

| Entry point | Location |
|---|---|
| `enqueueLinks` | `packages/core/src/enqueue_links/enqueue_links.ts` |
| `enqueueLinksByClickingElements` (Playwright) | `packages/playwright-crawler/src/internals/enqueue-links/click-elements.ts:208` |
| `enqueueLinksByClickingElements` (Puppeteer) | `packages/puppeteer-crawler/src/internals/enqueue-links/click-elements.ts` |
| `SitemapRequestLoader` | `packages/core/src/storages/sitemap_request_loader.ts:211` |

No deprecated options on `EnqueueLinksOptions`, no legacy types in `crawlee-core.api.md`, no `@apify/pseudo_url` dependency.

## Change 2 — `apify` SDK: `src/enqueue-links-v3-compat.ts`

Pseudocode:

```ts
import { Minimatch } from 'minimatch';
import { purlToRegExp } from '@apify/pseudo_url';

const OPTION_KEYS = ['method', 'payload', 'headers', 'userData', 'label'];
const LEGACY_KEYS = ['globs', 'regexps', 'pseudoUrls'];

// Build [{ match, pattern, opts }] from the three v3 arrays.
function buildEntries({ globs = [], regexps = [], pseudoUrls = [] }) {
    const entries = [];

    for (const item of globs) {
        const glob = (typeof item === 'string' ? item : item.glob)?.trim();
        if (!glob) continue;                                   // v3 skipped empty globs
        const m = new Minimatch(glob, { nocase: true });        // same options crawlee uses
        entries.push({ pattern: { glob }, match: (url) => m.match(url), opts: pick(item, OPTION_KEYS) });
    }

    for (const item of regexps) {
        const regexp = item instanceof RegExp ? item : item.regexp;
        entries.push({ pattern: { regexp }, match: (url) => regexp.test(url), opts: pick(item, OPTION_KEYS) });
    }

    for (const item of pseudoUrls) {
        const regexp = purlToRegExp(typeof item === 'string' ? item : item.purl);
        entries.push({ pattern: { regexp }, match: (url) => regexp.test(url), opts: pick(item, OPTION_KEYS) });
    }

    return entries;
}

/** Rewrites v3 enqueueLinks options into v4 shape. Returns the input untouched if no legacy keys. */
export function toV4EnqueueLinksOptions(options = {}) {
    if (!LEGACY_KEYS.some((k) => options[k]?.length)) return options;   // hot path: no-op

    const { globs, regexps, pseudoUrls, include, transformRequestFunction, ...rest } = options;
    const entries = buildEntries({ globs, regexps, pseudoUrls });

    log.deprecated('`globs`/`regexps`/`pseudoUrls` are deprecated — use `include`.');

    const merged = [...(include ?? []), ...entries.map((e) => e.pattern)];

    return {
        ...rest,
        include: merged.length ? merged : undefined,   // v4 rejects `include: []`
        transformRequestFunction: (req) => {
            Object.assign(req, entries.find((e) => e.match(req.url))?.opts);   // per-pattern options
            return transformRequestFunction ? transformRequestFunction(req) : req;
        },
    };
}
```

Registered once, in `Actor.init()`:

```ts
serviceLocator.setEnqueueLinksOptionsPreprocessor(toV4EnqueueLinksOptions);
```

`toV4EnqueueLinksOptions` stays exported — useful for tests and for converting options explicitly.

## Change 3 — types (SDK, same file)

The hook fixes runtime, not TypeScript: `enqueueLinks({ globs })` still fails to compile because `EnqueueLinksOptions` has no `globs`. It is an interface, so declaration merging works:

```ts
declare module '@crawlee/core' {
    interface EnqueueLinksOptions {
        /** @deprecated Use `include`. Handled by the Apify SDK compatibility layer. */
        globs?: (string | { glob: string } & LegacyRequestOptions)[];
        /** @deprecated Use `include`. */
        regexps?: (RegExp | { regexp: RegExp } & LegacyRequestOptions)[];
        /** @deprecated Use `include` with globs or regexps. */
        pseudoUrls?: (string | { purl: string } & LegacyRequestOptions)[];
    }
}
```

Caveat: the augmentation is global. It widens the type for every actor importing `apify`, including ones that never use the legacy keys. JavaScript actors are unaffected either way.

## What the actor writes

Nothing. Existing v3 code keeps working after the Crawlee bump:

```ts
import { CheerioCrawler } from 'crawlee';

await Actor.init();   // installs the preprocessor

const crawler = new CheerioCrawler({
    requestHandler: async ({ enqueueLinks }) => {
        await enqueueLinks({
            selector: input.linkSelector,
            globs: input.globs,          // incl. { glob, method: 'OPTIONS' } from the Console
            pseudoUrls: input.pseudoUrls,
            exclude: input.excludes,     // unchanged in v4, passes through untouched
        });
    },
});
```

## Knock-on effects

- **`apify/actor-templates`** — the six templates using `globs:` keep working on v4 unchanged; all of them call `Actor.init()` before `crawler.run()`. This removes the earlier constraint that the rename had to be atomic with the dependency bump. They should still be updated eventually: every run of a template-derived actor would print a deprecation warning, and templates are what people copy.

barjin:
Adds `createTransformRequestFunction()`, which turns the values of the `globs` and `pseudoUrls` input schema editors into a Crawlee v4 `transformRequestFunction` (skipping non-matching requests and applying the matched pattern's request options), together with the exported interfaces describing those input shapes. 

Once this is merged and the API docs are live, we want to replace the sections of the [Input Schema specification](https://docs.apify.com/actors/development/actor-definition/input-schema/specification/v1#array) mentioning Crawlee with links to the SDK docs.

Closes #687.


janbuchar (approved):

B4nan (approved):
pushed a few fixed found during staff review, hope you dont mind

barjin:
> pushed a few fixed found during staff review, hope you dont mind

No hard feelings, thanks! Any idea what's going on with the failing Node 26 test? It looks like https://github.com/microsoft/playwright/issues/40724, but that's afaik been solved (see the other commits in this PR).

B4nan:
An azure flake I guess, passed now just fine.

PR #694, opened 2026-08-20
fix: Remove storage purge() methods so that the storage frontends can fill in the functionality

janbuchar:
- counterpart to https://github.com/apify/crawlee/pull/4056#issue-comment-box

### DIFF
--- a/src/apify_key_value_store_backend.ts
+++ b/src/apify_key_value_store_backend.ts
@@ -18,7 +18,11 @@ import type { KeyValueStoreClient } from 'apify-client';
  * @internal
  */
 export class ApifyKeyValueStoreBackend implements KeyValueStoreBackend {
-    constructor(private readonly client: KeyValueStoreClient) {}
+    /** @param onDropped Lets the owning backend know the store is gone, so it never resolves to it again. */
+    constructor(
+        private readonly client: KeyValueStoreClient,
+        private readonly onDropped?: () => void,
+    ) {}
 
     async getMetadata(): Promise<KeyValueStoreInfo> {
         const metadata = await this.client.get();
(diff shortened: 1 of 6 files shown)


src/apify_key_value_store_backend.ts
```diff
@@ -18,7 +18,11 @@ import type { KeyValueStoreClient } from 'apify-client';
  * @internal
  */
 export class ApifyKeyValueStoreBackend implements KeyValueStoreBackend {
-    constructor(private readonly client: KeyValueStoreClient) {}
+    /** @param onDropped Lets the owning backend know the store is gone, so it never resolves to it again. */
+    constructor(
+        private readonly client: KeyValueStoreClient,
+        private readonly onDropped?: () => void,
```
barjin, line 24: Let's do the `private` / `#identifer` switch in SDK v4 as well?

src/apify_key_value_store_backend.ts
```diff
@@ -30,13 +34,26 @@ export class ApifyKeyValueStoreBackend implements KeyValueStoreBackend {
...
-        throw new Error(
-            'Purging a key-value store is not supported on the Apify platform. ' +
-                'Use `drop()` to delete the store entirely, or open a new store instead.',
-        );
+        let exclusiveStartKey: string | undefined;
+
+        do {
+            const { items, isTruncated, nextExclusiveStartKey } = await this.client.listKeys({ exclusiveStartKey });
+
+            // Sequentially: a store with many records would otherwise fire a whole page of deletes at once.
```
barjin, line 50: >  // Sequentially: a store with many records would otherwise fire a whole page of deletes at once.

We could have some reasonable-sized batches for this, but `perf` updates can always come later, this is strictly internal anyway.

src/apify_storage_backend.ts
```diff
@@ -181,6 +181,9 @@ export class ApifyStorageBackend implements StorageBackend {
     /** Fallback request queue client key when the run id is unavailable — one per backend. */
     private fallbackClientKey?: string;
 
+    /** Storages this backend dropped; the run's configuration and the Actor's schema still point at them. */
+    private readonly droppedStorageIds = new Set<string>();
```
barjin, line 185: What happens if

1. we drop a storage _(it's possible to call `.drop()` on its own, right?)_
2. the run migrates
3. we try to open the dropped storage

`droppedStorageIds` seems ephemeral (it might disappear with the migration), no?

barjin (commented):
Thank you @janbuchar ! 

A few ideas - the first two are nits, the last one (dropping & migration) I'm actually curious about.

janbuchar:
The tests fail because the crawlee part is not yet released.

src/apify_key_value_store_backend.ts
B4nan: #699

PR #700, opened 2026-08-21
refactor: Subscribe to crawlee status message events to restore Actor status message functionality

### LINKED ISSUE #696: Handle status message events from Crawlee
The event is emitted by crawlee BasicCrawler. This replaces the old `setStatusMessage` that piggybacked on the `StorageClient` interface (now `StorageBackend`).

janbuchar:
- closes #696

- crawlee doesn't call `StorageClient.setStatusMessage()` anymore (https://github.com/apify/crawlee/pull/3818), it broadcasts a `statusMessage` event instead, and nothing in the SDK listened - so crawler progress stopped reaching the run
- `Actor.init()` now forwards those events, without re-logging the message (the crawler already logged it, at its own level)
- `Actor.exit()` drops the listener after the drain, so a straggling periodic message can't overwrite the terminal status message
- the `crawlerId` property of the event is not utilized in any way at this point

### DIFF
--- a/src/actor.ts
+++ b/src/actor.ts
@@ -2,6 +2,7 @@ import { createPrivateKey } from 'node:crypto';
 
 import type {
     EventManager,
+    EventStatusMessageData,
     EventTypeName,
     IStorage,
     RecordOptions,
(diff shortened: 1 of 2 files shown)


barjin (approved):
👍 

----

nit: I see this as a rather internal refactor of how the status message in Crawlee / SDK works. Not sure this warrants a `fix:` Conventional Commits prefix.

PR #704, opened 2026-08-26
test: install musl-compatible native bindings in the e2e Actor image

vdusek:
`actorChargeLocal` and `multiStorageLocal` run the harness Actor against local storage, so they load Crawlee's Rust-backed filesystem storage. It fails on the platform with `Cannot find native binding`, which is what the last [E2E run](https://github.com/apify/apify-sdk-js/actions/runs/32945115384/job/98196640343) tripped over. The v4 line hasn't had a green E2E run since 2026-06-04, before that native package existed.

## Causes

- The harness image resolves its dependencies in `node:22` (Debian/glibc) and copies `node_modules` into `apify/actor-node:22` (Alpine/musl), so npm picks `@crawlee/fs-storage-native-linux-x64-gnu`. The first install also passed `--no-optional`, skipping the platform packages entirely.

## Changes

- The builder stage is `node:22-alpine`, matching the runtime image's libc.
- `--no-optional` is gone, and `--only=prod` becomes `--omit=dev` (the old spellings are deprecated and npm warned about them).
- `git` is installed in the builder. Alpine ships none, and pointing a dependency at a git branch is how the SDK gets tested against candidate `apify-client` builds (#702).
- `overrides` pins `@crawlee/fs-storage-native` to `0.1.5-beta.19`. The exact version matters: `0.1.5` sorts above `0.1.5-beta.19` but predates the musl targets. Drop the override once a Crawlee release picks that version up.

## Verification

Built the harness Actor image locally from this tree. The old Dockerfile reproduces the CI error verbatim; the new one installs `fs-storage-native-linux-x64-musl` and runs the charging path through `FileSystemStorageBackend`, producing the counts `actorChargeLocal` asserts — 4 events uncapped, 3 with `maxTotalChargeUsd: 2`.

## Follow-up

Crawlee should bump `@crawlee/fs-storage`'s native dependency to a musl-capable version. Until it does, Crawlee v4's filesystem storage cannot load on any `apify/actor-node` image.

*✍️ Drafted by Claude Code*


test/e2e/sdk/actorBase/package.json
```diff
@@ -9,6 +9,9 @@
     "dependencies": {
         "apify": "*"
     },
+    "overrides": {
+        "@crawlee/fs-storage-native": "^0.2.0"
```
janbuchar: Wait for https://github.com/apify/crawlee/pull/4077 and drop this

janbuchar (approved):
LGTM, but get rid of the overrides

test/e2e/sdk/actorBase/package.json
vdusek: Yeah, but then I have to bump the minimum version in `package.json` to:

```
"crawlee": ">=4.0.0-beta.151 <4.0.0-rc.0"
```

Is that OK?

janbuchar: Just to make sure I understand this. If we just keep `>=4.0.0-beta.151`, it will be resolved to `rc.0`, which pulls in the "old" version of `fs-storage-native`?

I'd rather hurry to make another RC if that's the case...

vdusek: Yes, exactly. Alright, I'll wait in this case.

PR #708, opened 2026-09-03
fix(storage): fix storage alias resolution

### LINKED ISSUE #697: resolveId throws on internal aliases minted by Crawlee
`ApifyStorageBackend.resolveId` resolves `__default__` from the run config and looks every other alias up in `ACTOR_STORAGES_JSON`, throwing when it is absent and `isAtHome`. Crawlee v4 mints two kinds of alias that will never be declared there:

- `__default_<n>__` — the second and subsequent `BasicCrawler` instances in one process, so they don't share the default request queue. This is a documented v4 feature.
- `throttled-<domain>` — one per domain, created lazily by `ThrottlingRequestManager` as the crawl discovers them, i.e. whenever `sameDomainDelaySecs` is set.

Both work locally, where the non-`isAtHome` branch creates an unnamed storage per alias. On the platform they throw:

> Storage alias "throttled-example.com" cannot be resolved because it is not declared in the Actor's schema storages. Declare it in the Actor schema, or open the storage by name instead.

Neither piece of advice is actionable: the aliases are generated at runtime, and `throttled-<domain>` depends on which domains the crawl happens to find. So `sameDomainDelaySecs`, and any Actor running two crawlers, are broken on the platform and only on the platform — which is why no test or local run catches it.

Two ways out:

- Apply the local fallback (create an unnamed storage) to aliases carrying a reserved prefix, so a user's typo still errors but Crawlee's internal aliases resolve.
- Or have Crawlee stop expressing run-scoped internal storages as user-space aliases at all. That needs a separate channel on `StorageBackend` and would land on the Crawlee side instead.

janbuchar:
- Fixed storage alias resolution handling in ApifyStorageBackend and storage opening.
- Updated storage exports and configuration handling.
- closes #697

### DIFF
--- a/src/actor.ts
+++ b/src/actor.ts
@@ -4,13 +4,11 @@ import type {
     EventManager,
     EventStatusMessageData,
     EventTypeName,
-    IStorage,
     RecordOptions,
-    StorageOpenOptions,
     UseStateOptions,
 } from '@crawlee/core';
 import { Dataset, EventType, KeyValueStore, purgeDefaultStorages, RequestQueue, serviceLocator } from '@crawlee/core';
-import type { Awaitable, Constructor, Dictionary, StorageBackend } from '@crawlee/types';
+import type { Awaitable, Dictionary, StorageBackend } from '@crawlee/types';
 import { sleep } from '@crawlee/utils';
 import type {
     ActorCallOptions,
@@ -50,8 +48,7 @@ import { getDefaultsFromInputSchema, noActorInputSchemaDefinedMarker, readInputS
 import { PlatformEventManager } from './platform_event_manager.js';
 import type { ProxyConfigurationOptions } from './proxy_configuration.js';
 import { ProxyConfiguration } from './proxy_configuration.js';
-import type { OpenStorageOptions, StorageIdentifier, StorageIdentifierWithoutAlias } from './storage.js';
-import { openStorage } from './storage.js';
+import type { OpenStorageOptions, StorageIdentifier } from './storage.js';
 import {
     checkCrawleeVersion,
     getSystemInfo,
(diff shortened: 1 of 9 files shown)


src/actor.ts
```diff
@@ -1176,7 +1166,9 @@ export class Actor<Data extends Dictionary = Dictionary> {
 
         this._ensureActorInit('openDataset');
 
-        return this._openStorage<Dataset<Data>>(Dataset, datasetIdOrName, options);
+        return Dataset.open<Data>(datasetIdOrName ?? null, {
+            storageBackend: options.forceCloud ? this.createApifyStorageBackend() : undefined,
```
B4nan, line 1170: what about doing this instead (and returning `undefined` from `createApifyStorageBackend` conditionally)

```suggestion
            storageBackend: this.createApifyStorageBackend(options.forceCloud),
```

src/apify_storage_backend.ts
```diff
@@ -296,22 +319,62 @@ export class ApifyStorageBackend implements StorageBackend {
...
+
+            const { id } = await this.collectionClient(type).getOrCreate();
+            this.aliasIdCache.set(key, id);
+            if (store) await this.persistAliasId(store, key, id);
+            return id;
+        });
+    }
+
+    /**
+     * Re-reads the record first, so a second backend in this process does not drop its entries.
```
barjin, line 361: > a second backend in this process does not drop its entries

If we have two separate backend instances, couldn't  $A_\text{read} B_\text{read} B_\text{write} A_\text{write}$ happen here (dropping the value written by $B$)?

Should we make the lock `static` so it applies to all the backend instances in one process?

barjin (approved):
Lgtm, just one potential parallelism issue / race condition ⬇️

src/apify_storage_backend.ts
janbuchar: Under normal circumstances, there should only ever be one backend instance that can manipulate the default key-value store. Due to how `forceCloud` is implemented in current `v4`, this is not the case. This will be resolved with https://github.com/apify/apify-sdk-js/issues/717

B4nan (commented):
of course i forgot to submit the review. i had just one nit, feel free to ignore

Issue #352, opened 2025-02-03
Possible race condition in PPE-enabled `Actor.pushData`


janbuchar:
FYI: By moving the push before the charge, we introduced a small race condition where as we are close to the max budget, more pushes can happen concurrently and we could push a few more items than we can charge for. I think this might happen maybe in 1/1000 runs, maybe even less (depends on the size of the data push which determines the latency of it).

I thought a bit about fixes for this and haven't come up with anything elegant. You could have a new `bookFutureCharge` method to pre-register events but it would introduce a new complex API. Or we could patch it on the `Actor` class bypassing the manager and just handle in the `pushData`. All of it introduces extra complexities. So for now I'm parking that thought because I don't think it is a big deal.

_Originally posted by @metalwarrior665 in https://github.com/apify/apify-sdk-js/pull/346#discussion_r1937993583_

PR #709, opened 2026-09-03
refactor: Simplify per-dataset-item PPE

janbuchar:
- related to #352 
- instead of subclassing `DatasetClient` from `apify-client`, we now do the synthetic `apify-default-dataset-item` accounting in `ChargingDatasetBackend`, a decorator installed on top of the `StorageBackend` by the `Actor` class
- the `AsyncLocalStorage` used to group together chunked `pushItems` calls is gone :rocket:
- the chunking logic in the Apify backend is intact

### DIFF
--- a/docs/02_concepts/08_pay_per_event.mdx
+++ b/docs/02_concepts/08_pay_per_event.mdx
@@ -60,13 +60,15 @@ The user spending limit for the run is available in your Actor code as the `ACTO
 
 When the charge limit is reached, <ApiLink to="class/Actor#charge">`Actor.charge`</ApiLink> stops charging and <ApiLink to="class/Actor#pushData">`Actor.pushData`</ApiLink> stops pushing data. The platform then aborts the run automatically. However, the run keeps consuming platform resources for a short time before it stops. For details, see [Handle graceful shutdown](https://docs.apify.com/platform/actors/publishing/monetize/pay-per-event#handle-graceful-shutdown).
 
-### Charging inside a request handler
+### Charge inside a request handler
 
-By default, crawlee runs each request handler in a storage transaction: dataset writes are buffered and applied when the handler finishes. Charging for dataset items follows the items, so a push inside a handler is billed at commit, and a handler that throws is retried without being billed for the items it discarded.
+By default, crawlee runs each request handler in a storage transaction: dataset writes are buffered and applied when the handler finishes. Charging for dataset items follows the items. If you push items in a handler, crawlee makes the charge at the commit. If the handler throws an error, crawlee retries the request. Crawlee doesn't charge for the items that it discarded.
 
-<ApiLink to="class/Actor#charge">`Actor.charge`</ApiLink> is not part of the transaction and takes effect immediately. It therefore consumes budget that dataset items buffered by the handler would otherwise be charged for - those items are dropped at commit if the budget no longer covers them.
+<ApiLink to="class/Actor#charge">`Actor.charge`</ApiLink> isn't part of the transaction and takes effect immediately. It therefore consumes budget that dataset items buffered by the handler would otherwise be charged for. If the budget isn't large enough to cover these items, crawlee discards them at the commit.
 
-For the same reason, <ApiLink to="class/Actor#pushData">`Actor.pushData`</ApiLink> with an event name throws inside a transaction: the charge cannot wait for the commit. Either wrap the call in crawlee's `withDirectStorageAccess()` to store and charge right away, accepting that a retried request pushes the items again, or push without an event name and charge separately.
+For the same reason, <ApiLink to="class/Actor#pushData">`Actor.pushData`</ApiLink> with an event name throws inside a transaction: the charge cannot wait for the commit. You have two options:
+- Wrap the call in crawlee's `withDirectStorageAccess()` function. It stores the items and charges immediately. However,  if crawlee retries the request, it pushes the items again.
+- Push the items without an event name. Then, make the charge separately.
 
 ## Test monetization locally
 
--- a/docs/02_concepts/08_pay_per_event.mdx
+++ b/docs/02_concepts/08_pay_per_event.mdx
@@ -9,6 +9,7 @@ import ConditionalActorChargeSource from '!!raw-loader!./code/conditional_actor_
 import ChargeLimitCheckSource from '!!raw-loader!./code/charge_limit_check.ts';
 import PerUnitChargingSource from '!!raw-loader!./code/per_unit_charging.ts';
 import ChargingManagerSource from '!!raw-loader!./code/charging_manager.ts';
+import ChargeInRequestHandlerSource from '!!raw-loader!./code/charge_in_request_handler.ts';
 import ApiLink from '@site/src/components/ApiLink';
 import CodeBlock from '@theme/CodeBlock';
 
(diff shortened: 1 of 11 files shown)


src/charging_storage_backend.ts
```diff
@@ -0,0 +1,139 @@
...
+            await this.inner.pushData(items);
+            return;
+        }
+
+        await chargingManager.withChargeLock(async () => {
+            const limit = chargingManager.calculatePushDataLimit(items.length, { isDefaultDataset: true });
+
+            if (limit === 0) {
+                return;
+            }
```
barjin, line 75: Btw, how does charging in `Dataset.pushData` now work with Crawlee transactions? The items are pushed + billed at the transaction commit? Are the "overcharged" items dropped?

What happens if we interleave explicit events and dataset items?

```typescript
await Actor.pushData([...]) // 100 items, $0.1 USD each. Will get billed only after the requestHandler run
const { eventChargeLimitReached } = await Actor.charge(...); // false, as we haven't billed the `pushData` events yet
```

Shall we document this somewhere (e.g., using direct access w/ PPE Actors as best practices)?

janbuchar: Check out dea88741a0 — I made it throw and ask the user to be explicit.

docs/02_concepts/08_pay_per_event.mdx
```diff
@@ -60,6 +60,14 @@ The user spending limit for the run is available in your Actor code as the `ACTO
 
 When the charge limit is reached, <ApiLink to="class/Actor#charge">`Actor.charge`</ApiLink> stops charging and <ApiLink to="class/Actor#pushData">`Actor.pushData`</ApiLink> stops pushing data. The platform then aborts the run automatically. However, the run keeps consuming platform resources for a short time before it stops. For details, see [Handle graceful shutdown](https://docs.apify.com/platform/actors/publishing/monetize/pay-per-event#handle-graceful-shutdown).
 
+### Charging inside a request handler
```
szaganek:
```suggestion
### Charge inside a request handler
```
Just a style guide rule, we avoid gerunds in headings.

docs/02_concepts/08_pay_per_event.mdx
```diff
@@ -60,6 +60,14 @@ The user spending limit for the run is available in your Actor code as the `ACTO
...
+
+By default, crawlee runs each request handler in a storage transaction: dataset writes are buffered and applied when the handler finishes. Charging for dataset items follows the items, so a push inside a handler is billed at commit, and a handler that throws is retried without being billed for the items it discarded.
```
szaganek:
```suggestion
By default, crawlee runs each request handler in a storage transaction: dataset writes are buffered and applied when the handler finishes. Charging for dataset items follows the items. If you push items in a handler, crawlee makes the charge at the commit. If the handler throws an error, crawlee retries the request. Crawlee doesn't charge for the items that it discarded.
```
How about making it a lil easier on the eyes like this.

docs/02_concepts/08_pay_per_event.mdx
```diff
@@ -60,6 +60,14 @@ The user spending limit for the run is available in your Actor code as the `ACTO
...
+
+<ApiLink to="class/Actor#charge">`Actor.charge`</ApiLink> is not part of the transaction and takes effect immediately. It therefore consumes budget that dataset items buffered by the handler would otherwise be charged for - those items are dropped at commit if the budget no longer covers them.
+
+For the same reason, <ApiLink to="class/Actor#pushData">`Actor.pushData`</ApiLink> with an event name throws inside a transaction: the charge cannot wait for the commit. Either wrap the call in crawlee's `withDirectStorageAccess()` to store and charge right away, accepting that a retried request pushes the items again, or push without an event name and charge separately.
```
szaganek:
```suggestion
For the same reason, <ApiLink to="class/Actor#pushData">`Actor.pushData`</ApiLink> with an event name throws inside a transaction: the charge cannot wait for the commit. You have two options:
- Wrap the call in crawlee's `withDirectStorageAccess()` function. It stores the items and charges immediately. However,  if crawlee retries the request, it pushes the items again.
- Push the items without an event name. Then, make the charge separately.
```
Turning the options into bullet points will make it easier to scan the info.

docs/02_concepts/08_pay_per_event.mdx
```diff
@@ -60,6 +60,14 @@ The user spending limit for the run is available in your Actor code as the `ACTO
 
 When the charge limit is reached, <ApiLink to="class/Actor#charge">`Actor.charge`</ApiLink> stops charging and <ApiLink to="class/Actor#pushData">`Actor.pushData`</ApiLink> stops pushing data. The platform then aborts the run automatically. However, the run keeps consuming platform resources for a short time before it stops. For details, see [Handle graceful shutdown](https://docs.apify.com/platform/actors/publishing/monetize/pay-per-event#handle-graceful-shutdown).
 
+### Charging inside a request handler
+
+By default, crawlee runs each request handler in a storage transaction: dataset writes are buffered and applied when the handler finishes. Charging for dataset items follows the items, so a push inside a handler is billed at commit, and a handler that throws is retried without being billed for the items it discarded.
+
+<ApiLink to="class/Actor#charge">`Actor.charge`</ApiLink> is not part of the transaction and takes effect immediately. It therefore consumes budget that dataset items buffered by the handler would otherwise be charged for - those items are dropped at commit if the budget no longer covers them.
```
szaganek:
```suggestion
<ApiLink to="class/Actor#charge">`Actor.charge`</ApiLink> isn't part of the transaction and takes effect immediately. It therefore consumes budget that dataset items buffered by the handler would otherwise be charged for. If the budget isn't large enough to cover these items, crawlee discards them at the commit.
```
Breaking a long sentence into two makes it easier to digest.

szaganek (approved):
A few rewrite suggestion from me :)

docs/02_concepts/08_pay_per_event.mdx
```diff
@@ -60,6 +60,16 @@ The user spending limit for the run is available in your Actor code as the `ACTO
...
 
+### Charge inside a request handler
+
+By default, crawlee runs each request handler in a storage transaction: dataset writes are buffered and applied when the handler finishes. Charging for dataset items follows the items. If you push items in a handler, crawlee makes the charge at the commit. If the handler throws an error, crawlee retries the request. Crawlee doesn't charge for the items that it discarded.
+
+<ApiLink to="class/Actor#charge">`Actor.charge`</ApiLink> isn't part of the transaction and takes effect immediately. It therefore consumes budget that dataset items buffered by the handler would otherwise be charged for. If the budget isn't large enough to cover these items, crawlee discards them at the commit.
+
+For the same reason, <ApiLink to="class/Actor#pushData">`Actor.pushData`</ApiLink> with an event name throws inside a transaction: the charge cannot wait for the commit. You have two options:
+- Wrap the call in crawlee's `withDirectStorageAccess()` function. It stores the items and charges immediately. However,  if crawlee retries the request, it pushes the items again.
+- Push the items without an event name. Then, make the charge separately.
```
barjin: > Then, make the charge separately

This sounds like the fix is as simple as this:

```typescript
async requestHandler() {
    ...
    await Actor.pushData([...]);
    await Actor.charge('event');
    ...
}
```

Which it isn't, as this code can double-charge now (on a request retry).

Perhaps `idempotencyKey` is ready to make its debut? 👀

barjin (approved):
just one small docs nit left ⬇️ 

Thanks @janbuchar !

docs/02_concepts/08_pay_per_event.mdx
janbuchar: [`d7a07b3` (this PR)](https://github.com/apify/apify-sdk-js/pull/709/commits/d7a07b3450d1f4a88c95e38288417fb41e9b5328)

PR #721, opened 2026-09-07
docs: document slim Docker images and package manager setup

vladfrangu:
Mirrors apify/crawlee#4082 for the *Running in Docker* guide, ported to the 3.7 snapshot as well:

- Document the `-slim` tag variants (no preinstalled `apify`/`crawlee`/`typescript`) and recommend them by default.
- New *Package managers* section: Corepack enabled (no package manager besides npm preinstalled), `packageManager` support, flat `node_modules` linker config, `/pkg-cache`, and how to override the linker via `ENV`.
- Supported Node.js versions are now 22, 24 and 26. Example Dockerfiles bumped from `:16` to `:24`.

[🤖] Claude Code using Fable 5


janbuchar (approved):
LGTM

PR #724, opened 2026-09-08
feat(charging): allow passing an idempotency key to charge()

ruocco-l:
Small addition to the charge method to allow passing the idempotency key (already allowed by the client)


barjin (approved):
Looks pretty straightforward 👍  Thanks @ruocco-l !

PR #727, opened 2026-09-10
ci: Remove beta release script

janbuchar:


.github/workflows/publish-to-npm.yaml
```diff
@@ -46,9 +46,26 @@ jobs:
             - name: Build module
               run: pnpm build
 
-            - name: Check version consistency and bump pre-release version (pre-release tags only)
+            # Only the pre-release version outputs are used here, so changelog generation and
+            # its GitHub API lookups are dead weight.
```
barjin, line 50: A new `generate_changelog: false` option for `git-cliff-release`?

barjin (approved):
lgtm, thanks!

.github/scripts/before-beta-release.js
```diff
@@ -1,64 +0,0 @@
...
-// ending in `v<major>` (e.g. `next-v4`) marks a pre-major release line:
-// package.json stays on the current major (3.x) and we publish betas of the
-// *next* major (4.0.0-beta.N) under that tag. (The tag isn't a bare `v4`
-// because npm refuses dist-tags that parse as a semver range.)
-// The `rc` tag works the same way but publishes release candidates of the
-// next major (e.g. 4.0.0-rc.N).
-const distTag = process.argv[2];
-const premajorMatch = /v(\d+)$/.exec(distTag ?? '');
-const premajor = premajorMatch ? Number(premajorMatch[1]) : null;
-const preid = distTag === 'rc' ? 'rc' : 'beta';
```
B4nan, line 21: so the release flow now handles everything including RC bumps and next major bumps as we have it in crawlee, right?

B4nan (approved):
lets adjust the docs around this a bit, not sure if the contributing guide here needs something, but we also have this outdated notion guide:

https://app.notion.com/p/apify/Release-Process-for-open-source-libs-7f2b48c4120b4ba6b9a7b3cae6c86583

.github/scripts/before-beta-release.js
janbuchar: This PR adds a pre-major git cliff config to this repo. It'd be cleaner to add this to `git-cliff-release` as well, I guess. I was hesitant to do it given how seldom we release majors, but let me whip that up real fast.

PR #731, opened 2026-09-11
feat: Add default storage id configuration options

janbuchar:
- Removed from crawlee (apify/crawlee#4125), re-introduced in SDK where they logically belong


barjin (approved):
lgtm, thanks!

PR #737, opened 2026-09-22
docs: stop repeating the section prefix in generated Markdown links

marekh19:
Part of apify/apify-web#6665

`@signalwire/docusaurus-plugin-llms-txt@1.2.2` prepends the site URL to link paths that already carry the Docusaurus `baseUrl`, so the generated Markdown points at `https://docs.apify.com/sdk/js/sdk/js/...`, which 404s. It only happens on sites served under a sub-path.

The fix is already upstream in `2.0.0-alpha.6`, but it does not look like it will be published soon: npm `latest` is still 1.2.2 and nothing has shipped since `2.0.0-alpha.7` in November 2025. This patch applies that same fix locally. Remove it once a stable 2.x lands.

Built here: `llms-full.txt` goes from 413 doubled links to 0, `llms.txt` stays at 91 correct links.

The same patch goes into `apify-cli`, `apify-sdk-python`, `apify-client-js` and `apify-client-python`.


patches/@signalwire__docusaurus-plugin-llms-txt@1.2.2.patch
```diff
@@ -0,0 +1,17 @@
...
+     if (relativePaths === false && baseUrl) {
+-        return normalizeUrl([baseUrl, targetPath]);
++        // baseUrl is the site URL plus the Docusaurus baseUrl, e.g. https://docs.apify.com/cli/.
++        // Some callers pass a routePath that already carries that baseUrl, so strip it before
++        // joining, or the section prefix ends up in the URL twice.
++        const basePath = baseUrl.replace(/^[a-z][a-z0-9+.-]*:\/\/[^/]*/i, '') || '/';
++        return normalizeUrl([baseUrl, stripBaseUrl(targetPath, basePath)]);
+     }
+     // For relative paths, ensure we preserve the leading slash
+     // normalizeUrl can sometimes remove it, so we handle it explicitly
```
B4nan: `.md` is appended before the base is stripped, so the site-root link (`/sdk/js/`, which is where the navbar "SDK for JavaScript" entry points) becomes `/sdk/js/.md`, then `/.md`, then `https://docs.apify.com/sdk/js/.md`. It is no longer doubled, so it is missing from the 413 to 0 count, but it still 404s. Line 137 of the current `llms-full.txt` has it.

Stripping right after `ensureLeadingSlash` fixes it: `/sdk/js/` collapses to `/`, which takes the `INDEX_MD` branch and ends up as `.../sdk/js/index.md`. I also cut the comment down to one line; the `/cli/` example came over from the apify-cli version of this patch.

I ran this ordering in a harness against both callers (`rehype-links` hrefs with and without the prefix, `tree-renderer` with `markdownFile`, `relativePaths: true`). Only the root route changes. Run `pnpm install` after applying so the lockfile hash is regenerated.

```suggestion
diff --git a/lib/utils/url.js b/lib/utils/url.js
index 0c3168c9128b6abd1a0860b222872b2332dc5ab8..995f88d3ee897a471c6aacdd4ae5a65aada599e2 100644
--- a/lib/utils/url.js
+++ b/lib/utils/url.js
@@ -58,6 +58,10 @@ export function formatUrl(routePath, options, baseUrl = '') {
     const { enableMarkdownFiles = true, relativePaths = true, markdownFile, } = options;
     // Ensure route path starts with /
     let targetPath = ensureLeadingSlash(routePath);
+    // routePath may already carry the Docusaurus baseUrl; strip it or the prefix appears twice.
+    if (relativePaths === false && baseUrl) {
+        targetPath = stripBaseUrl(targetPath, baseUrl.replace(/^[a-z][a-z0-9+.-]*:\/\/[^/]*/i, '') || '/');
+    }
     // Use markdown file path if available and enabled
     if (enableMarkdownFiles && markdownFile) {
         // Ensure markdown file path starts with / for consistency
```

package.json
```diff
@@ -129,6 +129,9 @@
             "fast-uri": "^3.1.2",
             "qs": "^6.15.2",
             "webpack-dev-server": "^6.0.0"
+        },
+        "patchedDependencies": {
+            "@signalwire/docusaurus-plugin-llms-txt@1.2.2": "patches/@signalwire__docusaurus-plugin-llms-txt@1.2.2.patch"
         }
```
B4nan: The other pnpm settings live in `pnpm-workspace.yaml`, and this file is the published `apify` manifest, so the key would ship to consumers (harmless, but noise). Move it there:

```yaml
patchedDependencies:
  '@signalwire/docusaurus-plugin-llms-txt@1.2.2': patches/@signalwire__docusaurus-plugin-llms-txt@1.2.2.patch
```

```suggestion
        }
```

B4nan (changes requested):

webrdaniel (approved):

B4nan (approved):

patches/@signalwire__docusaurus-plugin-llms-txt@1.2.2.patch
marekh19: @B4nan Good catch, thanks a lot! Also added the `'/sdk/js/',` to `excludedRoutes` otherwise the generated files link the site root at `https://docs.apify.com/sdk/js/index.md`, a file the build never produces, and the exclusion drops that dead `.md` target so the entry falls back to `https://docs.apify.com/sdk/js/`

Replicated this same setup across all repos.

